PR #4, opened 2023-06-20
Use the ratgdo pins configured in the substitutions section

mariusmuja:
Currently the `ratgdo` component in `base.yaml` always uses the default pin values. 

This results in compilation failure for boards that don't have the `Dx` pin aliases and doesn't allow for pins to be customized.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -8,6 +8,9 @@ external_components:
 
 ratgdo:
   id: ${id_prefix}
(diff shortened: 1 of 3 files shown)


bdraco (approved):

bdraco:
Will test this at the end of the week when I get home but I'm sure it's fine

PR #5, opened 2023-06-20
Add option for enable behaviour where light is triggered on by motion

mariusmuja:
In case of my garage opener the light turns on automatically when motion is detected, which causes the `ratdgo` light state to diverge from reality. This adds an option to enable the light to be switched on when motion is detected.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -8,6 +8,7 @@ external_components:
 
 ratgdo:
   id: ${id_prefix}
+  motion_triggers_light: ${motion_triggers_light}
 
 sensor:
   - platform: ratgdo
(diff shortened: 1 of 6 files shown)


bdraco:
When the motion activates I get a state update on all of mine that trigger the light.

I think a better solution would be to send the command to request state when the motion flips

bdraco:
One of the reboot commands triggers the gdo send back the current state. 

I can't remember off the top of my head which one it is though but it should be easy enough to test one at a time and than rename it.

When the motion event comes in we can fire off a write to pool state when it flips.  That should work for all known models

bdraco:
Probably only need to poll the state when it flips to on

mariusmuja:
Yes, I agree, firing up a status request is a nicer solution. I believe it's the REBOOT2 command, same as for the query button as that fixes the light being out of sync in my case.

PR #6, opened 2023-06-23
Local dev

mariusmuja:
This is not an actual pull request, I just wanted to make you aware of several improvements I made on my branch that you might be interested in merging. 

  * cleaned up/simplified the command handling
  * figured out that there are distinct ON/OFF/TOGGLE commands for light, lock and door open/close. This makes it less likely for state to get out of sync, for example when turning light off, sending an OFF command in place of TOGGLE will guarantee that light will be off. More importantly for closing the door, sending CLOSE command instead of TOGGLE. Door is somewhat peculiar, for example when opening it will ignore a CLOSE command (will not change direction), will only accept a TOGGLE command to stop. When closing however it will accept an OPEN command and will change direction (probably a safety feature). It's possible now however to open/stop/open some more/stop/open some more/...  which was not possible before.
  * sync() on startup was not working properly for me, probably because it contained blocking code (delay(..)). Fixed it by making delays async and realized that it's not necessary to send all those commands on startup, only sending GET_STATUS and GET_OPENINGS to sync state. On first use, the first command is ignored, so sending an extra GET_STATUS.
  * Speedup the serial read. A loop cycle in esphome happens once about every 16ms and an opener command on serial is transmitted in about 16ms as well, so by the time a byte can be read on the serial line the whole command can probably be read. Changed the code to read as much of the command as possible.
  * Save rolling code to flash once every 10s if it changed instead every some number of times it changed.
  * Removed all obstruction signal handing code (not necessary in my opinion) and use the opener obstruction state, which is what I care about after all. Makes the code simpler and allows for a simpler hw circuit.
  * Made remote_id a parameter of the component, which is seem it should be instead of hardcoded.


mariusmuja:
If you'd like I can open separate PRs for some of the above.

bdraco:
I took a look through it. It looks great.

Separate PRs for each point would be fantastic.

Thanks 👍

mariusmuja:
> * Removed all obstruction signal handing code (not necessary in my opinion) and use the opener obstruction state, which is what I care about after all. Makes the code simpler and allows for a simpler hw circuit.

The opener doesn't send a status update when the obstruction clears (need to keep querying for one), so I didn't create a PR for this change. The change is in commit ESPHome-RATGDO/esphome-ratgdo@f0b8380fa52e162c9fa200792725c5d1c3544e4f for future reference.

PR #8, opened 2023-06-23
Add auto close time sensor

bdraco:
If TTC is enabled, a sensor showing when the door is expected to auto close will be available.

Screenshot 2023-06-23 at 6 54 06 PM


base.yaml
```diff
@@ -3,6 +3,7 @@
 external_components:
   - source:
       type: git
+      ref: dev
```
```suggestion
      ref: auto_close_time
```

bdraco:
timestamp is wrong due to https://github.com/esphome/issues/issues/2875

bdraco:
This is probably DOA with the sensor platform because it only offers 7 digits of precision

bdraco:
closing due to https://github.com/ESPHome-RATGDO/esphome-ratgdo/pull/8#issuecomment-1605188281

PR #10, opened 2023-06-24
Refactor command handling

mariusmuja:
This PR cleans up/simplifies the command handling, adds support for distinct ON/OFF/TOGGLE commands for light, lock and door open/close. 

This makes it less likely for state to get out of sync, for example when turning light off, sending an OFF command in place of TOGGLE will guarantee that light will be off. More importantly for closing the door, sending CLOSE command instead of TOGGLE. Door is somewhat peculiar, for example when opening it will ignore a CLOSE command (will not change direction), will only accept a TOGGLE command to stop. When closing however it will accept an OPEN command and will change direction (probably a safety feature). It's possible now however to open/stop/open some more/stop/open some more/... which was not possible before.

Also added sending of GET_STATUS command when motion is detected and light is off (my opener doesn't send updated status on its own), and GET_OPENINGS command when door is closed to update the openings counter.

### DIFF
--- a/components/ratgdo/cover/ratgdo_cover.cpp
+++ b/components/ratgdo/cover/ratgdo_cover.cpp
@@ -26,12 +26,15 @@ namespace ratgdo {
             break;
         case DoorState::DOOR_STATE_OPENING:
             this->current_operation = COVER_OPERATION_OPENING;
+            this->position = 0.5;
             break;
         case DoorState::DOOR_STATE_CLOSING:
             this->current_operation = COVER_OPERATION_CLOSING;
+            this->position = 0.5;
             break;
         case DoorState::DOOR_STATE_STOPPED:
-            this->position = COVER_OPEN;
+            this->current_operation = COVER_OPERATION_IDLE;
+            this->position = 0.5;
         default:
             this->current_operation = COVER_OPERATION_IDLE;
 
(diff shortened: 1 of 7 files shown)


bdraco:
I've been formatting with

clang-format --style=Webkit -i  components/ratgdo/*.h components/ratgdo/*.cpp components/ratgdo/**/*.cpp components/ratgdo/**/*.h

mariusmuja:
Sounds good, I'll update formatting and rebase on main, looks like I used an older revision.

bdraco:
manual testing looks good

heap looks good

```
sensor:
# Uptime sensor.
- platform: template
  name: ratgdo Free Memory
  lambda: return ESP.getFreeHeap();
  unit_of_measurement: 'B'
  state_class: measurement
```
Screenshot 2023-06-24 at 3 33 12 PM

components/ratgdo/cover/ratgdo_cover.cpp
```diff
@@ -26,12 +26,15 @@ namespace ratgdo {
...
+            this->position = 0.5;
             break;
         case DoorState::DOOR_STATE_CLOSING:
             this->current_operation = COVER_OPERATION_CLOSING;
+            this->position = 0.5;
             break;
         case DoorState::DOOR_STATE_STOPPED:
-            this->position = COVER_OPEN;
+            this->current_operation = COVER_OPERATION_IDLE;
+            this->position = 0.5;
```
bdraco, line 37: This is a bit of an assumption but I think its the best we can do without knowing the timing of the garage open/close

bdraco (approved):

bdraco:
Thanks @mariusmuja

components/ratgdo/cover/ratgdo_cover.cpp
mariusmuja: Yes, what I would have liked is a 3 state cover type (OPEN/CLOSED/PARTIALLY OPEN). Since that's not available, I'm essentially assuming PARTIALLY OPEN is `position==0.5`.

mariusmuja:
Happy to contribute!

PR #11, opened 2023-06-24
Make remote_id a component parameter

mariusmuja:
Seems like remote_id should be a parameter of the component instead of hardcoded.


bdraco (approved):
Thanks @mariusmuja

PR #12, opened 2023-06-24
Use async delays in sync() and door open

mariusmuja:
It's not a good idea to use blocking `delay(..)` inside the `loop()` calls. Turn instances where `delay(..)` is used into async delays.

Also simplify `sync()` function, it doesn't seem to be necessary to send all those command on startup, only use GET_STATUS and GET_OPENINGS. Tested with a different remote_id to simulate first use and worked ok.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -21,6 +21,7 @@ namespace esphome {
 namespace ratgdo {
 
     static const char* const TAG = "ratgdo";
+    static const int SYNC_DELAY = 2000;
     static const uint8_t MAX_CODES_WITHOUT_FLASH_WRITE = 3;
 
     void IRAM_ATTR HOT RATGDOStore::isrObstruction(RATGDOStore* arg)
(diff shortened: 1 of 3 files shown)


components/ratgdo/ratgdo.cpp
```diff
@@ -21,7 +21,6 @@ namespace esphome {
 namespace ratgdo {
 
     static const char* const TAG = "ratgdo";
-    static const int STARTUP_DELAY = 2000; // delay before enabling interrupts
     static const uint8_t MAX_CODES_WITHOUT_FLASH_WRITE = 3;
```
bdraco, line 25: Is this unused now ?

mariusmuja: Seems like it's been unused for a while, I noticed it when checking for `delay` calls in the code and didn't find any place where it was used.

components/ratgdo/ratgdo.cpp
```diff
@@ -439,22 +438,24 @@ namespace ratgdo {
 
     void RATGDOComponent::sync()
     {
-        this->rollingCodeUpdatesEnabled_ = false;
-        for (int i = 0; i <= MAX_CODES_WITHOUT_FLASH_WRITE; i++) {
-            transmit(command::GET_OPENINGS); // get openings
-            delay(65);
```
bdraco: This was in there in case the esp lost power before it wrote the counter to ensure the counter never went backwards

mariusmuja: I see, I'll put that back in and update the PR.

bdraco:
Looks good, testing now

bdraco (approved):

PR #13, opened 2023-06-24
Rolling code flash sync updates

mariusmuja:
Save rolling code to flash every 10s in case it changed. Also perform flash sync outside of the component loop.


bdraco:
Looks good.

Testing now

bdraco:
rebooted during open
rebooted during close
reflashed during open
reflashed during close
pulled power during open
pulled power during close
everything seems to recover ok

bdraco (approved):

PR #14, opened 2023-06-24
Button updates

mariusmuja:
Replace button children with template buttons, since they don't need to sync any state, this seems simpler.
Also add "Query openings" button.

BREAKING CHANGE: The `Query` button has been replaced with `Query status`


base.yaml
```diff
@@ -157,15 +158,30 @@ button:
   - platform: safe_mode
     name: "Safe mode boot"
     entity_category: diagnostic
-  - platform: ratgdo
-    id: ${id_prefix}_sync
-    type: sync
+  
+  - platform: template
+    id: ${id_prefix}_query_status
```
bdraco, line 162:
```suggestion
    id: ${id_prefix}_query
```

bdraco: I guess we could avoid the breaking change, but I don't think its going to affect most people so maybe best to not carry it forward a legacy name

bdraco: Let's do the breaking change. This is too new to worry about it and its a diagnostic entity

bdraco (approved):

PR #20, opened 2023-06-25
Door position control

mariusmuja:
This implements door position control using the time it takes the door to open/close. It's obviously open-loop, since there's no feedback from the GDO until the door hits the fully opened / closed positions, it optimistically syncs the door position in the UI every 0.5s .

However, I find it quite useful, many times I want to open the garage door just a little (10-15%) to let some air in and now I can do it in one step instead of opening&stopping it.

It adds two configuration parameters, opening_duration & closing_duration, probably the same in most cases, but can be different if needed. If these are not configured, it assumes middle position 50% as before.

### DIFF
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -60,7 +60,6 @@ namespace ratgdo {
         } else if (this->number_type_ == RATGDO_CLOSING_DURATION) {
             this->parent_->setClosingDuration(value);
         }
-        // this->publish_state(value);
     }
 
 } // namespace ratgdo
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -10,17 +10,57 @@ namespace ratgdo {
     void RATGDONumber::dump_config()
     {
         LOG_NUMBER("", "RATGDO Number", this);
-        ESP_LOGCONFIG(TAG, "  Type: Rolling Code Counter");
+        if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
+            ESP_LOGCONFIG(TAG, "  Type: Rolling Code Counter");
+        } else if (this->number_type_ == RATGDO_OPENING_DURATION) {
+            ESP_LOGCONFIG(TAG, "  Type: Opening Duration");
+        } else if (this->number_type_ == RATGDO_CLOSING_DURATION) {
+            ESP_LOGCONFIG(TAG, "  Type: Closing Duration");
+        }
+    }
+
+    void RATGDONumber::set_number_type(NumberType number_type_) 
+    { 
+        this->number_type_ = number_type_; 
+        if (this->number_type_ == RATGDO_OPENING_DURATION || this->number_type_ == RATGDO_CLOSING_DURATION) {
+            this->traits.set_step(0.1);
+        }
     }
 
     void RATGDONumber::on_rolling_code_change(uint32_t rollingCodeCounter)
     {
+        if (this->number_type_ != RATGDO_ROLLING_CODE_COUNTER) {
+            return;
+        }
         this->publish_state(rollingCodeCounter);
     }
+
+    void RATGDONumber::on_opening_duration_change(float duration)
+    {
+        if (this->number_type_ != RATGDO_OPENING_DURATION) {
+            return;
+        }
+        this->publish_state(duration);
+    }
+
+    void RATGDONumber::on_closing_duration_change(float duration)
+    {
+        if (this->number_type_ != RATGDO_CLOSING_DURATION) {
+            return;
+        }
+        this->publish_state(duration);
+    }
+
     void RATGDONumber::control(float value)
     {
-        this->parent_->setRollingCodeCounter(value);
-        this->publish_state(value);
+        if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
+            this->parent_->setRollingCodeCounter(value);
+        } else if (this->number_type_ == RATGDO_OPENING_DURATION) {
+            this->parent_->setOpeningDuration(value);
+        } else if (this->number_type_ == RATGDO_CLOSING_DURATION) {
+            this->parent_->setClosingDuration(value);
+        }
+        // this->publish_state(value);
     }
 
 } // namespace ratgdo
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -170,25 +170,47 @@ namespace ratgdo {
         data &= ~0xf000; // clear parity nibble
 
         if ((fixed & 0xfff) == this->remote_id) { // my commands
-            ESP_LOGV(TAG, "[%ld] received mine: rolling=%07" PRIx32 " fixed=%010" PRIx64 " data=%08" PRIx32, millis(), rolling, fixed, data);
+            ESP_LOGD(TAG, "[%ld] received mine: rolling=%07" PRIx32 " fixed=%010" PRIx64 " data=%08" PRIx32, millis(), rolling, fixed, data);
             return 0;
         } else {
-            ESP_LOGV(TAG, "[%ld] received rolling=%07" PRIx32 " fixed=%010" PRIx64 " data=%08" PRIx32, millis(), rolling, fixed, data);
+            ESP_LOGD(TAG, "[%ld] received rolling=%07" PRIx32 " fixed=%010" PRIx64 " data=%08" PRIx32, millis(), rolling, fixed, data);
         }
 
         nibble = (data >> 8) & 0xff;
         byte1 = (data >> 16) & 0xff;
         byte2 = (data >> 24) & 0xff;
 
-        ESP_LOGV(TAG, "cmd=%03x (%s) byte2=%02x byte1=%02x nibble=%01x", cmd, cmd_name(cmd), byte2, byte1, nibble);
+        ESP_LOGD(TAG, "cmd=%03x (%s) byte2=%02x byte1=%02x nibble=%01x", cmd, cmd_name(cmd), byte2, byte1, nibble);
 
         if (cmd == command::STATUS) {
-            auto doorState = static_cast<DoorState>(nibble);
-            if (doorState == DoorState::DOOR_STATE_CLOSED && this->doorState != doorState) {
-                transmit(command::GET_OPENINGS);
+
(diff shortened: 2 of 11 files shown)


components/ratgdo/number/ratgdo_number.cpp
```diff
@@ -10,17 +10,57 @@ namespace ratgdo {
...
-        this->parent_->setRollingCodeCounter(value);
-        this->publish_state(value);
+        if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
+            this->parent_->setRollingCodeCounter(value);
+        } else if (this->number_type_ == RATGDO_OPENING_DURATION) {
+            this->parent_->setOpeningDuration(value);
+        } else if (this->number_type_ == RATGDO_CLOSING_DURATION) {
+            this->parent_->setClosingDuration(value);
+        }
+        // this->publish_state(value);
```
bdraco: Stale comment?

components/ratgdo/ratgdo.cpp
```diff
@@ -176,6 +189,29 @@ namespace ratgdo {
             }
```
bdraco, line 212: Maybe move the transmit to the end of this block so it happens after

components/ratgdo/ratgdo.cpp
```diff
@@ -46,10 +46,23 @@ namespace ratgdo {
...
+        this->rollingCodePref_ = global_preferences->make_preference<int>(734874333U);
+        if (!this->rollingCodePref_.load(&this->rollingCodeCounter)) {
             this->rollingCodeCounter = 0;
         }
+        this->openingDurationPref_ = global_preferences->make_preference<float>(734874334U);
+        if (!this->openingDurationPref_.load(&this->openingDuration)) {
+            this->setOpeningDuration(0);
+        } else {
+            this->sendOpeningDuration();
+        }
```
bdraco, line 58: Does this need to be configurable? can we calculate it from the maximum travel time from when the state changes from closed - opening - open (without stopping, or other command coming in) and save it?

components/ratgdo/ratgdo.cpp
```diff
@@ -232,11 +268,57 @@ namespace ratgdo {
...
     }
 
+    void RATGDOComponent::setOpeningDuration(float duration)
+    {
+        ESP_LOGD(TAG, "Set opening duration: %.1fs", duration);
+        this->openingDuration = duration;
+        this->openingDurationPref_.save(&this->openingDuration);
+        set_timeout("sync_preferences", 500, [=] {
+            global_preferences->sync();
+        });
```
bdraco: I think its best to let the 5s duration elapse and have preferences write it instead

components/ratgdo/ratgdo.cpp
```diff
@@ -232,11 +268,57 @@ namespace ratgdo {
...
+    }
+
+    void RATGDOComponent::setClosingDuration(float duration)
+    {
+        ESP_LOGD(TAG, "Set closing duration: %.1fs", duration);
+        this->closingDuration = duration;
+        this->closingDurationPref_.save(&this->closingDuration);
+        set_timeout("sync_preferences", 500, [=] {
+            global_preferences->sync();
+        });
```
bdraco: https://github.com/ESPHome-RATGDO/esphome-ratgdo/pull/20/files#r1241281018

components/ratgdo/number/ratgdo_number.cpp
mariusmuja: I comment it out because I thought it wasn't necessary and wanted to test. Is it actually needed here and under which conditions?

components/ratgdo/ratgdo.cpp
mariusmuja: That's a good idea! No more "calibration" necessary!

components/ratgdo/number/ratgdo_number.cpp
bdraco: I don't think it is needed. My comment was about a comment being left in

mariusmuja: Oh, I see. I didn't take it out completely because I wasn't 100% sure it wasn't needed.

bdraco: The client that changes should reflect it but I guess if you have the webserver open and hass it would need to be reflected back as the one that didn't change it probably isn't going to see the change otherwise

mariusmuja: Yes, I did wonder that too, but I tried with the webserver and hass opened and the change was reflected.

mariusmuja:
I added automatic open/close duration calibration. It keeps averaging in time with the current setting in case there's any drift.

I kept the settings for now. Could be useful in case manual reset of the durations is necessary for any reason (missed commands leading to erroneous duration being calculated?). Will see how it works...

I noticed you changed some logging levels from debug to verbose... can we keep them at debug? I don't want to set the global level to verbose (esphome recompiles everything with verbose level), but I pretty much always want to see those messages. You could set the ratgdo level to INFO if you don't want to see them. (I cannot set ratgdo level to VERBOSE while global level is DEBUG).

components/ratgdo/ratgdo.cpp
```diff
@@ -485,28 +595,127 @@ namespace ratgdo {
...
+            doorCommand(data::DOOR_CLOSE);
+            this->movingToPosition = true;
+            set_timeout("move_to_position", closing_time, [=] {
+                doorCommand(data::DOOR_STOP);
+                this->movingToPosition = false;
+                this->doorPosition = position;
+            });
+            this->positionSyncWhileClosing(delta);
+        }
+    }
```
bdraco, line 708: There is a bit of duplication here. It would be nice if this was a bit more DRY

components/ratgdo/ratgdo.cpp
```diff
@@ -157,31 +170,81 @@ namespace ratgdo {
...
+            }
+
+            if (this->doorState == DoorState::DOOR_STATE_OPENING && !this->movingToPosition) {
+                this->positionSyncWhileOpening(1.0 - this->doorPosition);
+                this->movingToPosition = true;
+            }
+            if (this->doorState == DoorState::DOOR_STATE_CLOSING && !this->movingToPosition) {
+                this->positionSyncWhileClosing(this->doorPosition);
+                this->movingToPosition = true;
+            }
```
bdraco, line 231:
```suggestion
            if (!this->movingToPosition) {
                if (this->doorState == DoorState::DOOR_STATE_OPENING) {
                    this->positionSyncWhileOpening(1.0 - this->doorPosition);
                } else if (this->doorState == DoorState::DOOR_STATE_CLOSING) {
                    this->positionSyncWhileClosing(this->doorPosition);
                }
                this->movingToPosition = true;
            }
```

one less conditional but doesn't really save that much

bdraco:
I got a crash right away on opening after installing this

mariusmuja:
Hmmm... I've been playing with it for a while without any crashes. It's likely that you hit a first use bug that I "avoided", can you send a log from before the crash?

bdraco:
After the crash everything works though so it might be because something was initially populated.

Will wipe a fresh one and try it with a console attached.

bdraco:
I didn't have the console attached when it crashed (requires a latter and a long usb)

mariusmuja:
> I didn't have the console attached when it crashed (requires a latter and a long usb)

I usually monitor the log over the network (`esphome logs gdo.yaml`), doesn't catch everything a serial console might, but pretty close.

bdraco:
> > I didn't have the console attached when it crashed (requires a latter and a long usb)
> 
> I usually monitor the log over the network (`esphome logs gdo.yaml`), doesn't catch everything a serial console might, but pretty close.

I had that open at the time but it just caught the connection reset not the crash dump.

I reflashed again and can't replicate it.

I think I'm going to have to clear NVS

bdraco:
Tried setting it to 0 and can't replicate the crash

Will flash the other GDO and have serial attached when I do the first open


Will take a bit to get the ladder setup  etc

components/ratgdo/ratgdo.cpp
mariusmuja: Indeed, but removing the duplication makes the code quite a bit less readable I think.

bdraco:
I've done 5 erases and it hasn't crashed

bdraco:
Large garage
Screenshot 2023-06-25 at 5 54 10 PM

Small garage
Screenshot 2023-06-25 at 5 54 10 PM


Both the same height but the large one is slightly heavier

mariusmuja:
Do those doors have exactly the same times, or did you use the same image twice?

Here's mine, don't know why it's in such a hurry :)

image

bdraco:
20 clear + flashes and I can't replicate the issue

components/ratgdo/ratgdo.cpp
```diff
@@ -46,10 +46,23 @@ namespace ratgdo {
 
     void RATGDOComponent::setup()
     {
-        this->pref_ = global_preferences->make_preference<int>(734874333U);
-        if (!this->pref_.load(&this->rollingCodeCounter)) {
+        this->rollingCodePref_ = global_preferences->make_preference<int>(734874333U);
+        if (!this->rollingCodePref_.load(&this->rollingCodeCounter)) {
             this->rollingCodeCounter = 0;
         }
+        this->openingDurationPref_ = global_preferences->make_preference<float>(734874334U);
```
mariusmuja, line 53: Just thinking about what could have caused the crash... Did you ever have a pref with the same type (734874334U) before while testing and it picked up a spurious duration?

bdraco: maybe thats it. 

I've flashed 3 other GDOs and no issues.

I think its fine. I'm going to merge it.

I'll see if I can make it a bit more DRY in a followup without making it less readable.

bdraco (approved):

components/ratgdo/ratgdo.cpp
bdraco: I flashed 7 in total. Only the one I've been doing the development on had the issue so I hope its a fluke

bdraco:
Thanks @mariusmuja  

This is awesome 🥇

components/ratgdo/ratgdo.cpp
bdraco: https://github.com/ESPHome-RATGDO/esphome-ratgdo/pull/21

I think `set_timeout` was being called with a negative value when closing

bdraco: nah I read it wrong when I refactored

PR #21, opened 2023-06-25
Small cleanup to make setDoorPosition more DRY

bdraco:


components/ratgdo/ratgdo.cpp
```diff
@@ -675,39 +676,36 @@ namespace ratgdo {
...
-            set_timeout("move_to_position", closing_time, [=] {
-                doorCommand(data::DOOR_STOP);
-                this->movingToPosition = false;
-                this->doorPosition = position;
-            });
-            this->positionSyncWhileClosing(delta);
+            this->positionSyncWhileClosing(-delta);
         }
+
+        auto operation_time = duration * 1000 * delta;
```
bdraco, line 699: I think on close `set_timeout` could have been called with a negative value before?

```
            delta = -delta;
            auto closing_time = this->closingDuration * 1000 * delta;
```

mariusmuja: Before it was in the `if (delta<0)` condition body, but in this PR seems like you're not accounting for that.

components/ratgdo/ratgdo.cpp
```diff
@@ -675,39 +676,36 @@ namespace ratgdo {
...
+        } else { // close
             doorCommand(data::DOOR_CLOSE);
-            this->movingToPosition = true;
-            set_timeout("move_to_position", closing_time, [=] {
-                doorCommand(data::DOOR_STOP);
-                this->movingToPosition = false;
-                this->doorPosition = position;
-            });
-            this->positionSyncWhileClosing(delta);
+            this->positionSyncWhileClosing(-delta);
```
bdraco: delta has to < 0 here because if its == 0 we already returned above

mariusmuja:
Looks good now!

mariusmuja (approved):

PR #22, opened 2023-06-26
Better sync

mariusmuja:
No need to send multiple GET_STATUS commands to increment rolling counter, we can just increment the rolling counter, the opener will accept forward jumps in the counter.

Improved the get status and openings by retrying to request the updates with backoff. My opener seems to stop sending replies for the same command sent multiple times in quick succession until some amount of time has passed.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
+      ref: better_sync
     refresh: 1s
 
 preferences:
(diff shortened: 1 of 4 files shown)


bdraco:
Looks good. Will test when I get back home in a few hours

bdraco:
I think it will accept up to 1000 jumps so we could make it even safer against unhandled resets by always bumping by 10

mariusmuja:
Sounds good to me.

bdraco (approved):
Manual testing is good 👍 

Thanks @mariusmuja

PR #24, opened 2023-06-26
Refactor children state update mechanism

mariusmuja:
This is a large-ish one... it refactors how state updates are pushed to the children from using callbacks in a common parent class to using an observer pattern.

 The members that need to be synced to the children are of an observable type now (ex. `observable<DoorState>`) and their state is automatically propagated to the children that subscribe to it. Also, the children components can subscribe only to the state they are interested in, so things like `if (this->binary_sensor_type_ != SensorType::RATGDO_SENSOR_OBSTRUCTION) return;` are no longer neccessary.

Also it's no longer necessary to keep the current and previous state values of everything and the `statusUpdateLoop()` is gone. Children updates are deferred until after the ratgdo component loop and multiple updates to the same observable within a loop only notify once. 

The ratgdo code remains unchanged with one difference: state is assigned to as usual (`this->doorState = ...`), but when it's value is needed it requires a dereference (`*this->doorState`).


mariusmuja:
That's strange: 
`ratgdo.cpp:577:26: error: use of 'auto' in lambda parameter declaration only available with -std=c++14 or -std=gnu++14`

My esphome (2023.5.5) seems to use gnu++17... I'll update the PR without auto in lamdas.

bdraco:
This makes sense. The dispatch than reject pattern was getting to be a bit much with the growing number of children

bdraco:
Setting as draft to pull it out of my review queue. Please set to ready when CI is passing 👍

mariusmuja:
I don't understand why CI is failing, seems like it's testing the wrong revision? The error message is:
```
src/esphome/components/ratgdo/ratgdo.cpp: In member function 'void esphome::ratgdo::RATGDOComponent::sync()':\nsrc/esphome/components/ratgdo/ratgdo.cpp:577:26: error: use of 'auto' in lambda parameter declaration only available with -std=c++14 or -std=gnu++14\n             300, 10, [=](auto r) 
```
There is no `auto` used in `RATGDOComponent::sync()` in the latest commit(s) (and `sync()` is not at line 577).

bdraco:
I think `main` is broken and the CI is building after that actually

bdraco:
I'll test this manually and merge it if it works

mariusmuja:
> I think `main` is broken and the CI is building after that actually

I see, yes that explains it, main has `auto` in the lamda  in `sync()`. This PR fixes that.

bdraco:
Free memory fell from 24888 to 20504

bdraco:
Everything seems to be working fine though. Would be nice if more could be moved back to flash as the esp8266 is getting dangerously low on ram

mariusmuja:
Hmm, it's been hovering around 24k on my esp8266 running with this PR for the last 2 days:

image

I think the jump from 24k to 26k in the graph is with/without `esphome logs`

mariusmuja:
I'm also using noise encryption, but I removed the dry sensors from my config as my hw circuit doesn't have them, probably that's the difference.

bdraco:
A few more bytes came back after waiting

Screenshot 2023-06-29 at 2 02 14 PM

bdraco:
I'll reflash to main and see if it comes back

bdraco:
Screenshot 2023-06-29 at 2 04 29 PM


baseline with this change

bdraco:
reflashed to main and baseline is about 24888 

Screenshot 2023-06-29 at 2 04 26 PM

bdraco:
Need to turn on noise and see how memory is as well

bdraco:
17344.0 B with noise turned on but it is stable.

Using more ram is going to be a problem past this point.

It would be great if we can come up with a way to shrink this

bdraco (approved):

bdraco:
The stalling I was seeing was actually caused by the logging. I downgraded a few to ESP_LOGV and everything seems stable now

mariusmuja:
> 17344.0 B with noise turned on but it is stable.

Mine has about 27k free (with noise encryption on)  and the difference from the 17k you're seeing has to be from the fact that I removed the dry contact/status gpios from the configuration. 

Out of curiosity, I wanted to see how it would behave with less free memory, so I added a dummy allocation to bring it down to about 5k. It seemed to work fine for me, the only issue I encountered was having to reboot in safe mode to do an OTA.

image

bdraco:
I had some instability only when using the web server with low memory. Likely because of the json memory allocation

mariusmuja:
> The stalling I was seeing was actually caused by the logging. I downgraded a few to ESP_LOGV and everything seems stable now

Interesting, I didn't know you were seeing any stalling issues. One thing that the default config could have is `baud_rate: 0` under the `logging:` component to disable UART logging. UART logging is useful in case of crases, but in such cases it can be re-flashed with UART logging on.

mariusmuja:
> I had some instability only when using the web server with low memory. Likely because of the json memory allocation

I see, that makes sense.

bdraco:
> > The stalling I was seeing was actually caused by the logging. I downgraded a few to ESP_LOGV and everything seems stable now
> 
> 
> 
> Interesting, I didn't know you were seeing any stalling issues. One thing that the default config could have is `baud_rate: 0` under the `logging:` component to disable UART logging. UART logging is useful in case of crases, but in such cases it can be re-flashed with UART logging on.

Everything seems fine at the moment with downgrading a few of them to verbose. That's good to keep in mind for the future if we need it

PR #27, opened 2023-07-02
Refactorings, changed from camelCase to snake_case for consistency to esphome

mariusmuja:
This renames all class members from camelCase to  snake_case for consistency to the esphome code convention.


bdraco:
Tested on ESP8266 and ESP32. All good

Thanks @mariusmuja

bdraco (approved):

PR #28, opened 2023-07-03
Convert enums to scoped enums

mariusmuja:
This converts enums to scoped enums (`enum class`) for better type safety. It also eliminates some code by adding an `ENUM()` macro that generates the enum together with an `EnumName_to_string(enum)` functios and a safe from int conversion `to_EnumName(int, default)`.

Also adds `this->` to member method calls as per the esphome coding conventions.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
+      ref: enums_refactor
     refresh: 1s
 
 preferences:
--- a/base.yaml
+++ b/base.yaml
@@ -4,7 +4,6 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
-      ref: enums_refactor
     refresh: 1s
 
 preferences:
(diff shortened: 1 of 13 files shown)


bdraco:
Side note. It look like memory allocation / collection is a bit more up and down after #27

I installed it around 10pm last night

Screenshot 2023-07-03 at 8 09 49 AM

mariusmuja:
Hmm, that's very strange, I don't see anything in #27 that could cause that. The only thing memory related in #27 is the move of `txRollingCode` and `txRollingCode` from class members to the stack & static variable, but that should not impact the heap.

Also, for me the memory profile has always looked like it looks for you now. Are you sure that #27 was the only change that happened (no esphome updates)?. I guess you could try reverting to before #27 and see if changes the memory allocation to how it was before.

bdraco:
No esphome update happened. I'll try reverting #27

bdraco:
I just looked at the graph again which is running this PR and its back to how it was before

Screenshot 2023-07-03 at 11 45 07 AM

bdraco (approved):

bdraco:
> Hmm, that's very strange, I don't see anything in #27 that could cause that. The only thing memory related in #27 is the move of `txRollingCode` and `txRollingCode` from class members to the stack & static variable, but that should not impact the heap.

Agreed. I'm not sure whats going on there. I'll keep watching it

Issue #33, opened 2023-07-20
Lock opening and closing duration?


DennisGarvey:
Sometimes the lock and closing duration change off of the values they are supposed to be. I cannot catch when they do it, but every time I have to set them to zero, fully open and close to door to recablibrate it. Is there a way to force those values to a certain number?

mariusmuja:
> Sometimes the lock and closing duration change off of the values they are supposed to be

Currently the open and close duration are continuously calibrating by averaging the exiting value with the latest opening/closing duration.

The only reason I can see for them to change (significantly) is if the opener doesn't always send a status message when the door has finished opening/closing and esphome-ratgdo mistakenly records an incorrect duration at some later point when a status message is eventually received. 

It might make sense to remove the continuous calibration in light of this issue and only calibrate them when they are zero.

bdraco:
Closing as this was implemented in https://github.com/ESPHome-RATGDO/esphome-ratgdo/pull/34

agelwarg:
> Is there any consideration of making these values non-editable by the user to avoid this confusion (unless I'm the only person who's tried this)?

I agree with @nwithan8 and was confused by the same thing.  I thought somehow it was maybe related to the TTC (Time To Close) topic that's been discussed, causing flashing/beeping when closing to warn anyone nearby.

PR #34, opened 2023-07-20
Don't calibrate open/close durations if they are non-zero

mariusmuja:
This is to fix issue #33, it only calibrates the opening&closing duration once, when they are zero.

In addition, it registers a timeout callback to request a status update 1 second after the door has opened/closed if one is not received (callback is cancelled if a status message is received before).

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -116,29 +116,31 @@ namespace ratgdo {
             auto door_state = to_DoorState(nibble, DoorState::UNKNOWN);
             auto prev_door_state = *this->door_state;
 
-            if (door_state == DoorState::OPENING && prev_door_state == DoorState::CLOSED) {
-                this->start_opening = millis();
-            }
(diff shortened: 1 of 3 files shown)


bdraco:
I think this is a good change 👍 

I'll test later today when I get back home

components/ratgdo/ratgdo.cpp
```diff
@@ -510,6 +504,12 @@ namespace ratgdo {
             this->door_position = *this->door_position + position_update;
             return RetryResult::RETRY;
         });
+
+        // this would only get called if no status message is received after door stops moving
+        // request a status message in that case, will get cancelled if a status message is received before
+        set_timeout("door_status_update", (*this->opening_duration + 1) * 1000, [=]() {
+            this->send_command(Command::GET_STATUS);
+        });
```
bdraco, line 512: nice!

bdraco:
Updated production. testing now

bdraco:
Did one full open/close cycle

All good.

bdraco (approved):
Thanks @mariusmuja 

This is a great change 👍

Issue #36, opened 2023-08-13
Obstruction Sensor Issue


mulcmu:
I have a brand new installation, everything is working except for the obstruction sensor feedback.  Installation was done using a `wemos_d1_mini32` as I plan to consolidate some other ESPhome devices currently in the garage into one node by the opener.  Brand new RATGDO v2 pcb and have tested with two separate ESP boards.

The obstruction sensor still is detected properly by the garage door opener.  I measure about 5.8V on the RATGDO terminal for obst/blk.  Meter measures frequency of 143 Hz which is right for pulses spaced 7ms apart on this terminal as well.  The RATGDO obst/blk terminal feeds the R1/R2 voltage divider into the D7 pin; continuity was confirmed so as far as I can tell the hardware setup should checkout.

I added some diagnostic log statements to the ISR routine and the obstruction loop.  `obsruction_low_count` remains at zero.  The log statement in the ISR never shows up in the log.

Any suggestions for additional troubleshooting?

```
    void IRAM_ATTR HOT RATGDOStore::isr_obstruction(RATGDOStore* arg)
    {
        // if (arg->input_obst.digital_read()) {
        //     arg->last_obstruction_high = millis();
        // } else {
        //     arg->obstruction_low_count++;
        // }

    arg->obstruction_low_count++;
    ESP_LOGD(TAG, "ISR");

    }

---------

    /*************************** OBSTRUCTION DETECTION ***************************/

    void RATGDOComponent::obstruction_loop()
    {
        long current_millis = millis();
        static unsigned long last_millis = 0;

        // the obstruction sensor has 3 states: clear (HIGH with LOW pulse every 7ms), obstructed (HIGH), asleep (LOW)
        // the transitions between awake and asleep are tricky because the voltage drops slowly when falling asleep
        // and is high without pulses when waking up

        // If at least 3 low pulses are counted within 50ms, the door is awake, not obstructed and we don't have to check anything else

        // Every 50ms
        if (current_millis - last_millis > 50) {
            // check to see if we got between 3 and 8 low pulses on the line
            if (this->isr_store_.obstruction_low_count >= 3 && this->isr_store_.obstruction_low_count <= 8) {
                // obstructionCleared();
                this->obstruction_state = ObstructionState::CLEAR;

                // if there have been no pulses the line is steady high or low
            } else if (this->isr_store_.obstruction_low_count == 0) {
[...]

PR #38, opened 2023-08-18
Fix obstruction isr for ESP32

mulcmu:
In RATGDOComponent::setup() call the ESPHome code to attach_interrupt() before software serial begin().  This allows ESPHome to call IDF gpio_install_isr_service() without error.

Tested on ESP32, fixes operation of the obstruction sensor from #36.  Still get an error in the logs presumably when software serial tries to call gpio_install_isr_service() during setup but connection to garage door works properly.

Needs to be tested on a ESP8266.


bdraco:
This looks fine but I'm traveling so I can't test

mariusmuja:
Tested and works ok on ESP8266.

PR #39, opened 2023-08-24
feat: add support for disabling the obstruction pin

bdraco:
Fall back to reading the obstruction state if the obstruction pin is disabled.

This is a subpar experience currently but better than nothing.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
+      ref: no_obst
     refresh: 1s
 
 preferences:
--- a/base.yaml
+++ b/base.yaml
@@ -4,7 +4,6 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
-      ref: no_obst
     refresh: 1s
 
 preferences:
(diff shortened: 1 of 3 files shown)


mariusmuja:
I implemented support for no obstruction pin at one point and the issue I was having was that there was not status message reliably sent when the instruction was cleared, so I had to ask for a status message in that case. Also there are two bits involved in reporting the obstruction, one gets cleared when an obstruction is present (byte1 - bit 6 - maybe a motor enabled bit?) and one gets set (byte2 - bit 2, obstruction detected bit?).  Sometimes a status message with the motor enabled bit set is sent, but no obstruction clear bit is sent.

See here the branch with these changes: https://github.com/ESPHome-RATGDO/esphome-ratgdo/commit/f0b8380fa52e162c9fa200792725c5d1c3544e4f#diff-8ec49aa77fc5c9af707edd16a08e03e3673aa20fc288a014c9d117f0edc5c2a3R159

bdraco:
> I implemented support for no obstruction pin at one point and the issue I was having was that there was not status message reliably sent when the instruction was cleared, so I had to ask for a status message in that case. Also there are two bits involved in reporting the obstruction, one gets cleared when an obstruction is present (byte1 - bit 6 - maybe a motor enabled bit?) and one gets set (byte2 - bit 2, obstruction detected bit?). Sometimes a status message with the motor enabled bit set is sent, but no obstruction clear bit is sent.

I found bit 6 is reliable (its a bit delayed) when the door is open but not when the door is closed

mariusmuja:
> I found bit 6 is reliable (its a bit delayed) when the door is open but not when the door is closed

Nice. I don't remember for sure, but it is quite possible I was testing with the door closed (but with the opener light on, otherwise the obstruction sensors go into sleep mode).

base.yaml
```diff
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
+      ref: no_obst
```
```suggestion
```

PR #41, opened 2023-08-26
Make obstruction detection ignore spurrious detections 

### LINKED ISSUE #40: Obstruction sensor triggers when turning on light
Hi --

First of all, fantastic job by all. I replaced my MyQ setup w/ this as it was so flaky and happy I've done it. The setup works stunningly, however, I did notice that when I turn _on_ my garage light, it briefly triggers a problem on the obstruction sensor (which then reverts back to OK). I didn't see this as an existing issue, not sure how to troubleshoot, but, my obstruction sensors work correctly when I manually trigger them, for what it's worth. Screengrab movie attached to demonstrate. I'm happy to try and help debug if you can tell me how to best log, etc.

https://github.com/ESPHome-RATGDO/esphome-ratgdo/assets/3329021/ed2e6bfa-b891-4c8f-bbf4-ca52371d1ba6

mariusmuja:
Make obstruction detection ignore spurrious detections when sensors awake from sleep mode, for example when light is turned on.

Fixes #40.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
+      ref: obstruction_detection
     refresh: 1s
 
 preferences:
--- a/base.yaml
+++ b/base.yaml
@@ -4,7 +4,6 @@ external_components:
   - source:
       type: git
       url: https://github.com/esphome-ratgdo/esphome-ratgdo
-      ref: obstruction_detection
     refresh: 1s
 
 preferences:
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -308,16 +308,16 @@ namespace ratgdo {
         // If at least 3 low pulses are counted within 50ms, the door is awake, not obstructed and we don't have to check anything else
 
         const long CHECK_PERIOD = 50;
-        const long PULSES_EXPECTED = CHECK_PERIOD/7;
+        const long PULSES_LOWER_LIMIT = 3;
 
         if (current_millis - last_millis > CHECK_PERIOD) {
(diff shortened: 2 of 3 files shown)


mulcmu:
I tested these on my setup.  The spurious detections were pretty intermittent before hand.  No spurious detections observed after the install and the obstruction sensor was still responsive to actual obstruction.

bdraco:
Looks good

Testing now

bdraco:
This does fix the light issue, but after this PR the obstruction sensor never turns off once it turns on for my gdos 

I may have fumbled the merge conflict though

mariusmuja:
I rebased on main, @bdraco maybe you can test now and see if the issue persists. If it does, we might need to relax the constraint on the number of pulses seen in order to clear the obstruction. Enable the debug here (https://github.com/ESPHome-RATGDO/esphome-ratgdo/blob/obstruction_detection/components/ratgdo/ratgdo.cpp#L314) and let me know the values you're seeing when the obstruction is not clearing.

bdraco:
Just got back home. testing now

bdraco:
https://hastebin.com/share/asuxibugul.yaml

once an obstruction is detected, time since asleep seems to increase forever

mulcmu:
The changes from this branch were working okay for me.  Here was debug output from the new obstruction loop:

[ratgdo log](https://github.com/ESPHome-RATGDO/esphome-ratgdo/files/12446372/ratgdo.log)

mariusmuja:
@bdraco Hmm, interesting. Your gdo obstruction sensors seems to clamp the voltage every 2ms-ish instead of 7ms. I removed the upper limit for the number of pulses requited to clear an obstruction, should take case of the issue.

bdraco:
Great. I'll be back home in an hour and give it another try

bdraco:
Works great. Super responsive now.  Will cleanup and merge

bdraco (approved):

bdraco:
Thanks @mariusmuja

great work 👍

PR #50, opened 2023-09-18
Refactor sync function to be more DRY and easier to extend

mariusmuja:
Small update that refactors `RATGDOComponent::sync()` to make it more DRY and easier to extend. 

Also, it only increments the rolling counter with `MAX_CODES_WITHOUT_FLASH_WRITE` if it doesn't receive a reply on first attempt.


bdraco:
Will try to test today before I leave (will be traveling until Oct 5th otherwise)

bdraco (approved):

Issue #52, opened 2023-09-28
Question/Request: Support for WEMOS LOLIN S2 Mini


rako77:
Basically I have some WEMOS LOLIN S2 Mini boards and I would like to use them with the ratgdo shield. I'm willing to try and get it working myself but I don't really understand how to compile this project?

Any pointers or tips?

mariusmuja:
Here's some some quick steps:

* Install `esphome`: https://esphome.io/guides/getting_started_command_line
* Checkout the project and modify `v2board_esp32_d1_mini.yaml` by changing `board: wemos_d1_mini32` to `board: lolin_s2_mini`.
* To compile the project run: `esphome compile v2board_esp32_d1_mini.yaml`
* To compile and upload to the board run: `esphome run v2board_esp32_d1_mini.yaml`

rako77:
Thank you!

TL;DR that helped. would you be interested in a pull request that adds support for the s2?

so far I have compiled and flashed the board. (the web flasher didn't work, I had to use: https://community.home-assistant.io/t/esp32-s2-and-esphome-upload/382885/18) I was able to connect the board to home assistant and set up the wifi using the web flasher.

todo: my ratgdo just shipped and hasn't arrived yet. So testing still hasn't happened.

Presuming it all functions as expected and given the caveat that the first flash might not work with the web flasher. would you like a PR?

bdraco:
> TL;DR that helped. would you be interested in a pull request that adds support for the s2?

Yes

mattjacobson6:
I'm still waiting on my board (October backorder), but I was also thinking of using the S2 Mini.  I assume there wouldn't be any issues with using the inner row of pinouts to connect other sensors (e.g., a PIR sensor)?  

Let us know how the S2 works with the Ratgdo!

PR #54, opened 2023-10-02
Lolin s2 mini support

rako77:
This PR aims to add support for the Lolin s2 min (https://www.wemos.cc/en/latest/s2/s2_mini.html)

It has been tested and works with my garage door.

As described in issue #52

Caveat:
Flashing from the web installer doesn't work for me however the web interface does allow for setting up the wifi once the board is flashed.
To flash manually I followed theses steps: https://community.home-assistant.io/t/esp32-s2-and-esphome-upload/382885/18


bdraco (approved):
Lgtm

bdraco:
Thanks @rako77

bdraco:
Looks like the image link needs to be fixed up as it's broken on the web installer page

rako77:
> Looks like the image link needs to be fixed up as it's broken on the web installer page

I'll submit a fix soon

PR #60, opened 2023-10-05
Allow interrupts to be disabled during transmit

mariusmuja:
By default in espsoftwareserial interrupts are enabled during transmit, so we receive back what we transmit on the wire. Worse than that, because we use the same line to TX & RX, the transmit operation gets interrupted to record the line transition times for RX decoding.

This change sets a flag that allows the interrupts to be disabled during TX. With this we no longer receive back what we transmit and no longer waste time decoding and handing that frame.

Additionally it tightens up the delays in the line transitions before transmitting to what seems to be necessary (together with an explanation comment on why I think it's like that).


bdraco (approved):
Testing 👍 

Nice improvement

Excellent work @mariusmuja

components/ratgdo/ratgdo.cpp
```diff
@@ -420,15 +412,17 @@ namespace ratgdo {
...
-                                                    // door opener responds to the message
-        delayMicroseconds(1305);
-        this->output_gdo_pin_->digital_write(false); // bring the line low
+        // indicate the start of a frame by pulling the 12V line low for at leat 1 byte followed by
+        // one STOP bit, which indicates to the receiving end that the start of the message follows
+        // The output pin is controlling a transistor, so the logic is inverted
+        this->output_gdo_pin_->digital_write(true); // pull the line low for at least 1 byte
+        delayMicroseconds(1300);
+        this->output_gdo_pin_->digital_write(false); // line high for at least 1 bit
+        delayMicroseconds(130);
```
bdraco, line 424: I wonder if this will cause a problem on some of the older openers....  If so we can always adjust it if someone reports an problem

mariusmuja: Yes, exactly. As long as the openers use 9600 baud, I don't think they should have an issue.

PR #61, opened 2023-10-07
Rework of door position sync

mariusmuja:
This updates the door position sync to make it simpler and hopefully more accurate.

The way we show the door position in HA is by "simulating it", because we don't have any position feedback from the door opener. Currently we simulate it and hope for the best, but for example if there's anything that prevents the open/close command to be received the simulation will not match reality.

This updates the simulation to make it slightly more realistic, instead to triggering it when we send the open/close command we trigger it when the opening/closing status message is received from the opener. Also we use the actual time in the update closure instead of the scheduled time.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -525,7 +525,6 @@ namespace ratgdo {
         if (*this->door_state == DoorState::OPENING) {
             return; // gets ignored by opener
         }
-        this->cancel_position_sync_callbacks();
 
         this->door_command(data::DOOR_OPEN);
     }
@@ -535,7 +534,6 @@ namespace ratgdo {
         if (*this->door_state == DoorState::CLOSING || *this->door_state == DoorState::OPENING) {
             return; // gets ignored by opener
         }
-        this->cancel_position_sync_callbacks();
 
         this->door_command(data::DOOR_CLOSE);
     }
@@ -554,7 +552,6 @@ namespace ratgdo {
         if (*this->door_state == DoorState::OPENING) {
             return; // gets ignored by opener
         }
-        this->cancel_position_sync_callbacks();
 
         this->door_command(data::DOOR_TOGGLE);
     }
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -264,12 +264,11 @@ namespace ratgdo {
 
     void RATGDOComponent::schedule_door_position_sync(float update_period)
     {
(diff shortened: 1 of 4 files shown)


components/ratgdo/ratgdo.cpp
```diff
@@ -16,6 +16,9 @@
 
 #include "esphome/core/log.h"
 
+#define ESP_LOG1 ESP_LOGV
+#define ESP_LOG2 ESP_LOGV
```
mariusmuja, line 20: I added these to have more flexibility on the logging output while debugging, instead of changing in a bunch of places `ESP_LOGV` -> `ESP_LOGD`, I can just change one of both of these to `ESP_LOGD` .

components/ratgdo/ratgdo.cpp
```diff
@@ -144,29 +146,67 @@ namespace ratgdo {
...
+                // request a status message in that case
+                set_timeout("door_status_update", (*this->opening_duration + 1) * 1000, [=]() {
+                    this->send_command(Command::GET_STATUS);
+                });
             }
-            if (door_state == DoorState::CLOSING && !this->moving_to_position) {
-                this->position_sync_while_closing(*this->door_position);
-                this->moving_to_position = true;
+
+            if (door_state == DoorState::CLOSING) {
```
bdraco: Looks like these could be else if blocks

mariusmuja: Oh, yeah, I was planing to update them to `else if`...

components/ratgdo/ratgdo.cpp
```diff
@@ -266,8 +266,8 @@ namespace ratgdo {
     {
         ESP_LOGD(TAG, "Schedule position sync: delta %f, start position: %f, start moving: %d",
             this->door_move_delta, this->door_start_position, this->door_start_moving);
-        auto duration = this->door_move_delta > 0 ? *this->opening_duration : -*this->closing_duration;
-        auto count = int(1000 * duration * this->door_move_delta / update_period);
+        auto duration = this->door_move_delta > 0 ? *this->opening_duration : *this->closing_duration;        
+        auto count = int(1000 * duration / update_period);
```
mariusmuja, line 270: I realized that it's better to schedule the door sync closure for the full open/close duration and have it cancelled when target is reached. Otherwise, if for example the door stop command fails, the position in HA stops at some percentage while the door fully closes or opens. This way the position in HA would be consistent with the actual door position.

bdraco:
Looks good.

Testing shortly

bdraco:
Ran out of time to test today. Will try to pick it up tomorrow

bdraco (approved):
Testing looks great 👍 

Nice cleanup 👍

PR #64, opened 2023-10-17
Door opening/close relliability updates

mariusmuja:
This adds two callbacks for command sent and door status received to ensure more predictable sequencing of events. 

For example, one case I encountered was when sending a door command (which is emulating a button press and a button release 100ms later), simultaneously with some motion events that caused the button press to be delayed, causing it to be too close or after the button release and thus ignored. This add a callback for when a door command was actually sent and only sends the next command then (instead of a fixed delay).

The door state callback is useful for things like sending a door command and waiting for a status message until another door command can be sent. This PR adds an `ensure_door_command` which can be used to ensure a command is actually received by the opener, otherwise it re-sends it after a certain delay.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,7 +4,6 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: reliability_updates
     refresh: 1s
 
 preferences:
--- a/base.yaml
+++ b/base.yaml
@@ -211,3 +211,11 @@ button:
       then:
         lambda: !lambda |-
           id($id_prefix).sync();
+
+  - platform: template
+    id: ${id_prefix}_toggle_door
+    name: "Toggle door"
+    on_press: 
+      then:
+        lambda: !lambda |-
+          id($id_prefix).toggle_door();
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
+      ref: reliability_updates
     refresh: 1s
 
 preferences:
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -620,9 +620,22 @@ namespace ratgdo {
     }
 
     void RATGDOComponent::ensure_door_command(uint32_t data, uint32_t delay)
-    {        
+    {
+        if (data == data::DOOR_TOGGLE) {
+            ESP_LOGW(TAG, "It's not recommended to use ensure_door_command with non-idempotent commands such as DOOR_TOGGLE");
+        }
+        auto prev_door_state = *this->door_state;
         this->door_state_received.then([=](DoorState s) {
-            // TODO: check door state is the expected one for command data?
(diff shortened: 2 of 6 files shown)


components/ratgdo/ratgdo.cpp
```diff
@@ -159,12 +159,6 @@ namespace ratgdo {
                     this->door_move_delta = 1.0 - this->door_start_position;
                 }
                 this->schedule_door_position_sync();
-
```
mariusmuja, line 162: This is replaced by `ensure_door_command`.

components/ratgdo/ratgdo.cpp
```diff
@@ -603,10 +611,25 @@ namespace ratgdo {
...
+                auto data2 = data & ~(1 << 8); // button release
+                this->send_command(Command::DOOR_ACTION, data2);
+            });
+        });
+    }
+
+    void RATGDOComponent::ensure_door_command(uint32_t data, uint32_t delay)
+    {        
+        this->door_state_received.then([=](DoorState s) {
+            // TODO: check door state is the expected one for command data?
```
bdraco: Is there still more to do here or is this ready for review?

mariusmuja: The TODO comment was there because I was undecided if further checking of the door state after was warranted... I ended up adding it. Should be ready for review now.

base.yaml
```diff
@@ -210,4 +210,12 @@ button:
...
\ No newline at end of file
+          id($id_prefix).sync();
+
+  - platform: template
+    id: ${id_prefix}_toggle_door
+    name: "Toggle door"
+    on_press: 
+      then:
+        lambda: !lambda |-
+          id($id_prefix).toggle_door();
```
mariusmuja, line 221: For a long time I thought that HA `cover.toggle` service ends up calling esphome's cover toggle (`call.get_toggle()` set to true) if the cover traits has `set_supports_toggle(true)`. It turns out that's not the case, so I added a button for people that want simple one-button control of the garage door form HA.

components/ratgdo/ratgdo.cpp
```diff
@@ -603,10 +611,38 @@ namespace ratgdo {
...
+        this->door_state_received.then([=](DoorState s) {
+            if ((data == data::DOOR_STOP) && (s != DoorState::STOPPED) && !(prev_door_state == DoorState::OPENING && s == DoorState::OPEN) && !(prev_door_state == DoorState::CLOSING && s == DoorState::CLOSED)) {
+                return;
+            }
+            if (data == data::DOOR_OPEN && !(s == DoorState::OPENING || s == DoorState::OPEN)) {
+                return;
+            }
+            if (data == data::DOOR_CLOSE && !(s == DoorState::CLOSED || s == DoorState::CLOSING)) {
+                return;
+            }
```
bdraco, line 637: This was a bit hard to follow bit the logic looks sound

bdraco:
I'll try to test this later today.  I'm on GMT-10 at the moment

bdraco (approved):
Thanks @mariusmuja , tested and working as expected.

Tested the new toggle as well 👍

PR #76, opened 2023-10-25
Use random client_id in lieu of static remote_id

bdraco:
## Breaking change

`remote_id` has been removed from YAML and is now a number entity called `Client ID`. Please remove it from your configuration if it has been specified manually.

## Proposed change

It is no longer necessary to save the rolling code counter when switching between firmware.

~~TODO~~
- [x] Test esp32
- [x] test esp8226
- [x] remove testing refs

### DIFF
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -61,13 +61,9 @@ namespace ratgdo {
             this->traits.set_step(0.1);
             this->traits.set_min_value(0.0);
             this->traits.set_max_value(180.0);
-        }
-        elif (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER)
-        {
+        } else if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
             this->traits.set_max_value(0xfffffff);
-        }
-        elif (this->number_type_ == RATGDO_CLIENT_ID)
-        {
+        } else if (this->number_type_ == RATGDO_CLIENT_ID) {
             // not sure how large remote_id can be, assuming not more than 24 bits
             this->traits.set_max_value(0xffffff);
         }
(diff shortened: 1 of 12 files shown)


components/ratgdo/number/ratgdo_number.cpp
```diff
@@ -65,7 +75,9 @@ namespace ratgdo {
 
     void RATGDONumber::control(float value)
     {
-        if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
+        if (this->number_type_ == RATGDO_CLIENT_ID) {
+            this->parent_->set_rolling_code_counter(value);
```
bdraco:
```suggestion
            this->parent_->set_client_id(value);
```

components/ratgdo/number/ratgdo_number.cpp
```diff
@@ -24,12 +26,20 @@ namespace ratgdo {
...
+            }
         }
         this->publish_state(value);
         this->control(value);
 
-        if (this->number_type_ == RATGDO_ROLLING_CODE_COUNTER) {
+        if (this->number_type_ == RATGDO_CLIENT_ID) {
+            this->parent_->subscribe_client_id([=](uint64_t value) {
+                this->update_state(value);
+            });
```
bdraco: Not sure we need this since it can't change from the main code path

components/ratgdo/ratgdo.h
```diff
@@ -133,8 +134,6 @@ namespace ratgdo {
         void set_output_gdo_pin(InternalGPIOPin* pin) { this->output_gdo_pin_ = pin; }
         void set_input_gdo_pin(InternalGPIOPin* pin) { this->input_gdo_pin_ = pin; }
         void set_input_obst_pin(InternalGPIOPin* pin) { this->input_obst_pin_ = pin; }
-        void set_remote_id(uint64_t remote_id) { this->remote_id_ = remote_id & 0xffffff; } // not sure how large remote_id can be, assuming not more than 24 bits
-        uint64_t get_remote_id() { return this->remote_id_; }
```
bdraco, line 137: I think we can keep this pattern since the only place it can change is the number entity

kbx81 (approved):
Working on my setup

mariusmuja:
Looks good :+1:

bdraco:
Thanks everyone for the testing and review.

Going to merge now

PR #81, opened 2023-10-26
Adding cover on_opening and on_closing triggers

### LINKED ISSUE #80: Request: "Door Status" on sooner while opening rather than open?
I use the door status output to turn on parking lasers that help with parking the cars in the right spot. I often enter the garage before the door is even fully up so the laser turns on a bit too late. It would be nice if the door status changed sooner as soon as the door starts opening. If you think about it, whether the door is partially or fully open, it is open and someone can enter. If there are use cases that would require that status to be high only when the door is fully open, then it would be nice for this to be configurable.

I am also happy to modify the code on my end as I need to modify it for other things such as being able to turn on the parking lasers manually when needed (mostly troubleshooting). I have also run into issues with the naming of entities (mostly the door as I want it to just use the name of the device since I called it "Overhead Garage Door") which led me to not use the 'package' feature.

mariusmuja:
This adds `on_opening` and `on_closing` triggers for the cover component, fixes #80.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -4,7 +4,6 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: cover_opening_closing_triggers
     refresh: 1s
 
 preferences:
--- a/base.yaml
+++ b/base.yaml
@@ -4,6 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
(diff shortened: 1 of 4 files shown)


bdraco:
Looks good.  I assume it's going to be just fine but I should actually test it to be sure 👍

bdraco (approved):
Thanks @mariusmuja

bdraco:
I forgot to merge this

PR #88, opened 2023-11-05
Fix door requiring two close commands to close

bdraco:
reported at https://github.com/home-assistant/core/issues/103336#issuecomment-1793613688 and confirmed locally

PR #92, opened 2023-11-05
Fix issue #89 by imposing some limits on the random client_id

### LINKED ISSUE #89: Sometimes the first close command is not accepted
https://github.com/ratgdo/esphome-ratgdo/pull/88 added a work around for this problem but it would be nice to discover the root cause and revert the work around.

I didn't have enough time to bisect far enough back to find where the problem was introduced and I didn't want to leave the issue in the wild so I settled on the work around

mariusmuja:
This should fix #89 (and reverts #88). It enforces the client_id to always end with `0x539`, but randomizes the two bytes before that. As an added benefit, it's easy to recognize the esp client_id (always ends in `0x539`).

Also fixes a bug in ratgdo_number where the component's state was not updated when the number value was changed (for example changing the client_id and refreshing/re-opening the web page or the HA device info page would show the old client_id, although the flash and ratgdo used the new one).

### DIFF
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -27,7 +27,7 @@ namespace ratgdo {
         this->pref_ = global_preferences->make_preference<float>(this->get_object_id_hash());
         if (!this->pref_.load(&value)) {
             if (this->number_type_ == RATGDO_CLIENT_ID) {
-                value = random(0x1, 0xFFFF);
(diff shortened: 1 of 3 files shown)


bdraco (approved):

bdraco:
> It enforces the client_id to always end with `0x539`

Great idea 👍

mariusmuja:
Don't merge yet, I'm about to commit a fix for those client_ids that are currently wrong.

mariusmuja:
Never mind... I'll commit as another PR.

bdraco:
Sorry too fast

PR #93, opened 2023-11-05
Update potentially wrong client_ids

mariusmuja:
Ensure existing ratgdo client_ids end with `0x539`.

### DIFF
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -36,8 +36,7 @@ namespace ratgdo {
             if (this->number_type_ == RATGDO_CLIENT_ID) {
                 uint32_t int_value = static_cast<uint32_t>(value);
                 if ((int_value & 0xFFF) != 0x539) {
-                    int_value = (int_value & ~0xFFF) | 0x539;
-                    value = int_value;
+                    value = ((random_uint32() + 1) % 0xFFFF) << 12 | 0x539;
                     this->pref_.save(&value);
                 }
             }
--- a/components/ratgdo/number/ratgdo_number.cpp
+++ b/components/ratgdo/number/ratgdo_number.cpp
@@ -32,6 +32,16 @@ namespace ratgdo {
                 value = 0;
             }
         }
+        else {
+            if (this->number_type_ == RATGDO_CLIENT_ID) {
+                uint32_t int_value = static_cast<uint32_t>(value);
+                if ((int_value & 0xFFF) != 0x539) {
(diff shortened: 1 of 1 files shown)


bdraco:
Will test shortly.  I still have my flasher connected so can test much more quickly.  I will also test a remote in service one

components/ratgdo/number/ratgdo_number.cpp
```diff
@@ -32,6 +32,16 @@ namespace ratgdo {
                 value = 0;
             }
         }
+        else {
+            if (this->number_type_ == RATGDO_CLIENT_ID) {
+                uint32_t int_value = static_cast<uint32_t>(value);
+                if ((int_value & 0xFFF) != 0x539) {
+                    int_value = (int_value & ~0xFFF) | 0x539;
+                    value = int_value;
+                    this->pref_.save(&value);
```
bdraco: Every one of them ended up being `1337` after reflashing

mariusmuja: Yeah, that's expected is they were less than `0xfff` (4095) before flashing (`1337`=`0x539`).

bdraco: Maybe it makes sense to generate a new one using `value = ((random_uint32() + 1) % 0xFFFF) << 12 | 0x539;` if it gets past  `if ((int_value & 0xFFF) != 0x539) {`?

That should ensure everyone gets a new working id and they don't end up back at 1337 which might be an issue if they switch firmwares in case the counter is too far behind?

mariusmuja: Sure.

bdraco:
Set it to 42, reflashed, got 39064890.0

bdraco (approved):
Everything working 👍

bdraco:
Thanks @mariusmuja

bdraco:
Screenshot 2023-11-05 at 6 13 26 PM

Just noticed we need to increase the max.  Will PR it in a moment

mariusmuja:
Yeah, just noticed that too...

mariusmuja:
Oh, the number sensor is not a good fit for this setting (unless we restrict the size of the client_id). Currently it's 28 bits (2 bytes  followed by 0x539) and that cannot always be represented by a float precisely (because not enough bits in the fraction part).

bdraco:
I think its fine to restrict the size of the client id. It just needs to be random enough that its unlikely they hit a collision when switching between firmware types

jfroy:
I can confirm my client ID was also automatically updated. 👍

PR #94, opened 2023-11-06
Increase max client id to 0xFFFFFFFF

bdraco:
to match the increase in #92


bdraco:
tested on a few production setups . all good. 

No more red box in HA

PR #95, opened 2023-11-06
Add limits on the range of client_id values

mariusmuja:
Restrict the range of client_ids so that we don't get conversion errors due to https://github.com/esphome/issues/issues/2875


bdraco (approved):

bdraco:
Thanks @mariusmuja

bdraco:
@jfroy the workarounds have been removed. We should no longer generate a bad id. When you get a chance, can you retest to make sure everything still works?

jfroy:
> @jfroy the workarounds have been removed. We should no longer generate a bad id. When you get a chance, can you retest to make sure everything still works?

@bdraco Will do.

jfroy:
I can confirm the GDO seems responsive with a single service call for open and close. I checked that my client ID has been updated to a valid bit pattern.

PR #97, opened 2023-11-07
Normalize client_id when manually entered

mariusmuja:
This updates/normalizes the client_id when manually entered in the number box.


bdraco (approved):
tested, thanks @mariusmuja

PR #104, opened 2023-11-08
Add timeout for motion detection

### LINKED ISSUE #103: motion detector not really working correctly.
The motion detector that is attached to my wall correctly triggers motion ON in the ratgdo logs.  It doesn't, however, turn off until it receives a 'OFF' which is essentially the 'light timer' embedded in the wall timing out.  The issue here though is it detects motion quite consistently.  We can see here in my logs:

```

17:34:50 | [D] | [ratgdo:246] | Motion: DETECTED
-- | -- | -- | --
17:34:52 | [D] | [binary_sensor:036] | 'Motion': Sending state ON
17:34:56 | [D] | [ratgdo:246] | Motion: DETECTED
17:35:00 | [D] | [ratgdo:246] | Motion: DETECTED
17:35:06 | [D] | [ratgdo:246] | Motion: DETECTED
17:38:11 | [D] | [ratgdo:210] | Status: door=CLOSED light=OFF lock=UNLOCKED
17:38:11 | [D] | [binary_sensor:036] | 'Motion': Sending state OFF

```

Since the motion detector is quite capable of faster intervals of detection it would be good if the associated home assistant device would trigger off far faster than waiting for the light to turn off which sends the OFF motion state. It looks like it detects motion down to 5 seconds or less.  Since it doesn't send the off though till the light goes off, its worthless in HA because we wont get a motion detected state change less than ever 3-5 minutes.

Can we set this to just automatically turn off at 3 second intervals or something so the  Motion: DETECTED will be sent again?

mariusmuja:
This adds a 3s timeout for the motion detection sensor, fixes #103.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -240,6 +240,9 @@ namespace ratgdo {
             }
         } else if (cmd == Command::MOTION) {
             this->motion_state = MotionState::DETECTED;
+            this->set_timeout("clear_motion", 3000, [=] {
+                this->motion_state = MotionState::CLEAR;
+            });
             if (*this->light_state == LightState::OFF) {
                 this->send_command(Command::GET_STATUS);
             }


components/ratgdo/ratgdo.cpp
```diff
@@ -240,6 +240,9 @@ namespace ratgdo {
             }
         } else if (cmd == Command::MOTION) {
             this->motion_state = MotionState::DETECTED;
+            this->set_timeout("clear_motion", 3000, [=] {
+                this->motion_state = MotionState::CLEAR;
+            });
```
bdraco, line 245: Is it possible that it could get detected again within 3s?  If so should the timer be canceled and rescheduled?

mariusmuja: > canceled and rescheduled?

It is, due to the `"clear_motion"` label.

bdraco: Right, of course https://github.com/esphome/esphome/blob/3e3266fa744dc234883d7f9aa8837e27f74bef3f/esphome/core/scheduler.cpp#L27

bdraco (approved):

Issue #111, opened 2023-11-16
Log spammed with `Collision detected, waiting to send packet`


ptr727:
With the device tethered for programming, but not connected to the opener yet, the log gets spammed with thousands of `Collision detected, waiting to send packet` events, that result in the esphome dashboard becoming non-responsive.

Code is [here](https://github.com/ratgdo/esphome-ratgdo/blob/e36225cfcee9ae7bdb164afceb97d7e0b15634b0/components/ratgdo/ratgdo.cpp#L467)

Maybe some preamble is required to either prevent the high rate of logs or wait for input to be connected.

Using a ratgdo v2.5 with a D1 mini.

bdraco:
duplicate of https://github.com/ratgdo/esphome-ratgdo/issues/47

Its not designed to function without being connected to the GDO

PR #143, opened 2023-12-16
Detect "not connected to GDO" condition

mariusmuja:
After some time (5s) of TX line being in not ready state, change "collision detected" message to "not connected to GDO" when a packet is sent.

Also, no longer floods log with "collision detected" messages, fixing #111.


bdraco:
Tested for regressions only. I didn't try it not connected.

bdraco (approved):

bdraco:
Thanks @mariusmuja

PR #144, opened 2023-12-16
Drop an incomplete packet if the missing bytes are lost

mariusmuja:
Currently if a packet is only partially received for any reason (noise on the line, flaky mosfet, ...), the next packet will be missed as well. 

For example if only 18 out of 19 bytes of packet `A` are read, when packet `A+1` comes, the first byte (`0x55`) will complete packet `A` (which in turn will fail to decode) and the rest of the bytes of packet `A+1` will be discarded because it will be starting with `0x01 0x00`, not the expected `0x55 0x01 0x00` header.

This change drops a packet if some time passes (100ms) with the packet still incomplete (a packet should be received in ~20ms). It logs a warning when a packet is dropped, since it should not occur under normal operation.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -318,8 +318,7 @@ namespace ratgdo {
 
     void RATGDOComponent::print_packet(const WirePacket& packet) const
     {
-        ESP_LOGV(TAG, "Counter: %d Send code: [%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X%02X]",
-            *this->rolling_code_counter,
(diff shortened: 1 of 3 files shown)


components/ratgdo/ratgdo.cpp
```diff
@@ -417,16 +419,28 @@ namespace ratgdo {
...
                     byte_count = 0;
+                    this->print_packet(rx_packet);
                     this->decode_packet(rx_packet);
                     return;
                 }
             }
+
+            if (millis() - last_read > 100) {
+                // if we have a partial packet and it's been over 100ms since last byte was read,
+                // the rest is not comming (a full packet should be received in ~20ms),
```
bdraco:
```suggestion
                // the rest is not coming (a full packet should be received in ~20ms),
```

bdraco:
Nice!

Testing this now

bdraco (approved):
Thanks @mariusmuja 

This makes a lot of sense as its inevitable that it will happen at some point. 👍

PR #190, opened 2024-01-21
Reduce number of flash writes

bdraco:
5s is a bit aggressive, and since we have a solution for catching up, lets set this back to 1m


mariusmuja:
I think it's a good idea, I've been using 1min all along on my GDO.

PR #194, opened 2024-01-23
Fix bug with opening/closing duration calibration

mariusmuja:
This fixes a bug recently introduced in main that breaks calibration of opening/closing duration.


bdraco (approved):
LGTM. Did not test because the change is trivial.

bdraco (approved):
Tested, all good

PR #196, opened 2024-01-23
feat: ensure all components in base.yaml have an id

### LINKED ISSUE #106: Include ID's for customization
Very, very, very low priority request.... 

Most entities include an ID, and while esphome is a bit flakey in parsing substituted !extend values, this sort of setting seems to work and I have maybe a dozen of them, for entities I do not seem to need in Home Assistant (and are still visible in the web server): 

    button: 
      - id: !extend ${id_prefix}_toggle_door
        internal: true

Two entities do not have them: restart and safe_mode_boot.  Maybe that's it, but if practical could all entities have id's so one can extend them?

brgaulin:
fixes #106

Example usage of these ids:

```yaml
packages:
  remote_package:
    files: [base.yaml]
    refresh: 1s
button: 
  - id: !remove ${id_prefix}_safe_mode
  - id: !extend ${id_prefix}_restart
    internal: true
```


brgaulin:
The changes I made for `api:` seemed to cause issues. I removed those and was able to remove the service with the following:
```yaml
api:
  services: !remove
```

bdraco (approved):

bdraco:
It looks like https://github.com/ratgdo/esphome-ratgdo/blob/main/base_secplusv1.yaml needs the same change

brgaulin:
Good catch! Added in the base_secplusv1.yaml ids

bdraco (approved):

bdraco:
Thanks @brgaulin

PR #200, opened 2024-01-23
New implementation of firmware selector

PaulWieland:
Two step selection process. First for protocol, second for ratgdo hw version.

image


static/index.html
```diff
@@ -312,19 +277,38 @@ <h3>Advanced Users</h3>
...
+      }
+
+      if(protocol === "secplusv2"){
+        protocol = "";
+      }else{
+        protocol = `_${protocol}`;
+      }
+
+      button.manifest = `${hardware}${protocol}-manifest.json`;
+      console.log(button.manifest);
```
bdraco: Left over console log

bdraco:
Are there 2.5 esp32s?

PaulWieland:
> Are there 2.5 esp32s?

Unknown. There are a tiny percentage of v2 shields and an equally tiny percentage of v2.5 shields, none of which were sold with an esp32.

static/index.html
PaulWieland: Yeah, made it easier to verify that the logic was producing the right manifest src. Will remove.

bdraco (approved):

bdraco:
Oops, looks like we are missing the image for the v2.5 esp32

PR #208, opened 2024-01-26
Fix overlapping firmware filenames

### LINKED ISSUE #209: ESPHome install page pointing to incorrect firmware file?
@mariusmuja oddly I flashed the ESPHome firmware with Security 1.0 to my RATGDO 2.5i board; however the log entries reference sec2plus ([ratgdo_secplus2:xxx]). Is the ESPHome install page pointing to incorrect firmware file?

_Originally posted by @miles267 in https://github.com/ratgdo/esphome-ratgdo/issues/128#issuecomment-1912405253_

bdraco:
The name of the firmware files were not unique enough for the builder as it uses the config name. To fix this I have forked the builder and made the following change: https://github.com/ratgdo/esphome-build-action/commit/61acb0e89cb5a47e9106d24861a7018a27907add

I also added the D1 mini (non lite) 8266 for the v2.5 board

fixes #209


PaulWieland:
reported in #209

bdraco:
changing the name is going to break mdns so we can't do that

PR #212, opened 2024-01-29
Change board in v25board_esp8266_d1_mini.yaml to d1_mini

ScottEgan:
I think static/v25board_esp8266_d1_mini.yaml needs a board definition update to reflect that the board is d1_mini now.


bdraco (approved):

bdraco:
thanks

PR #247, opened 2024-03-19
remove garage from cover entity, so devices can be named consistently

dlgoodr:
as it stands, if you call your ratgdo "Little Garage", you end up with a cover named `cover.little_garage_garage_door`, and a light named `light.little_garage_light`

After this change, you'd have `cover.little_garage_door` and `light.little_garage_light`, which feels more internally correct.

Conversely if you name a ratgdo "Little" to avoid the "double garage" in your cover entity, you'd end up with `cover.little_garage_door` and `light.little_light`. The light entity could be troublesome - call your garage "Main" and you suddenly have a "main_light" which probably overstates its importance in your environment.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -142,7 +142,7 @@ binary_sensor:
           condition:
             binary_sensor.is_off: ${id_prefix}_dry_contact_close
           then:
-            - cover.open: ${id_prefix}_garage_door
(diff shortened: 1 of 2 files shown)


dlgoodr:
Withdrawn.

PR #249, opened 2024-04-01
Initial commit to make the component aware of the dry_contact sensors…

PaulWieland:
…  and eliminate the need for lamda calls


mariusmuja:
@PaulWieland Have a look at this branch https://github.com/ratgdo/esphome-ratgdo/tree/dry_contact_with_sensors

PR #281, opened 2024-05-27
base: add uptime sensor

visigoth:
An uptime sensor is a nice-to-have debugging facility as it lets me verify that ratgdo is getting consistent power when using a voltage regulator connected to the battery backup terminals of my garage door opener.


visigoth:
I tested this with `esphome run v25iboard.yaml` and saw the uptime sensor in the table on the device's web page.

bdraco:
Uptime sensors are generally not enabled by default and usually only for debugging as they create many state updates and fill up Home Assistant's database quickly.

visigoth:
didn't know that; is there a way to have HA ignore it?

ccutrer:
You could set `disabled_by_default: true` in the ESPHome YAML, then a user would have to explicitly enable it in Home Assistant.

bdraco:
@visigoth Thanks for your PR. Because of the concern above https://github.com/ratgdo/esphome-ratgdo/pull/281#issuecomment-2132604433 , I think we should not add this to the default config.  Instead it would be nice to have a documentation PR on how to add it manually.

Thanks

Issue #214, opened 2024-01-30
Sec+ 1 & 2, Dry contact door toggle


PaulWieland:
When both trigger open & trigger close are pulled low together, ratgdo should toggle the door.

This feature doesn't currently work in the ESP Home firmware.

ColColonCleaner:
This would be great to have!

Jefficus:
Do I understand correctly that this would let me wire up a basic doorbell button to use as a secondary trigger without having to pay through the nose for a fancy digital push-button? If so, then I'll be eagerly watching this space for breaking announcements.

PaulWieland:
Yes thats what the dry contact inputs are for. There are discrete open and close dry contact inputs, so you need two momentary buttons at the moment until this change is implemented.

saschawj:
Are there any updates to this from the ESPHome developers? Is it something that will be worked on or is it going to be ignored?

mattjm:
So what's the "correct" way to go about implementing this?  As far as I can tell ESPHome doesn't have a "trigger on two binary sensors" function built-in.  I made a first pass at [using lambdas and global variables](https://github.com/mattjm/esphome-ratgdo/blob/9f08a209434138cc0086f0f55aa2134041feb6bf/base.yaml#L144) to run the toggle function when both buttons had been pressed, while preserving the single button function.  It's not currently working but if this is the completely wrong approach I'd rather direct my efforts at a smarter way.

saschawj:
Now that dry contacts have been implemented, is there a configuration setting for triggering the toggle?

mattjm:
There's no configuration or implementation for toggle via dry contact yet, as far as I know.  This issue is to update the firmware such that if you wire one side of your switch to both contacts, pushing the button will toggle the door (so really the "configuration" is in hardware 😂 ).  

I abandoned the lambda approach I posted earlier, and I finally [got something working](https://github.com/mattjm/esphome-ratgdo/blob/1f3ea8ae129426f99fcdbba651ff4f0d6f4fd414/base.yaml#L129), but I need to test it a while longer. There is a question of appropriate values for delayed_on and delayed_off--I've changed them to be lower here since my dry contact "button" is actually electronically triggered (it's an old system that uses a separate transmitter/receiver to trigger the door using a dry contact--specifically the "flash2pass" system marketed to motorcycles).  

[...]

PR #284, opened 2024-06-02
#214 -- toggle door when open and close dry contacts are triggered together

mattjm:
This addresses #214.  I've been testing it for about a week on a 25i board and it seems reliable.  

One thing to note is I made the on-state delay shorter for the buttons--I have a device that is triggering the contacts electronically and it doesn't pull them low long enough for the 500ms delayed_on.  I suspect other folks using this feature may be in a similar situation where they have some kind of legacy device they want to use alongside ratgdo.


PaulWieland:
Thanks for the PR. I already implemented this feature (without realizing this PR was made).  

https://github.com/ratgdo/esphome-ratgdo/pull/307/files

The simple `else` statement should suffice, but if not, please let me know.

PR #291, opened 2024-06-20
ota: add esphome platform

flavio-fernandes:
Starting on version 2024.6.0, at least one platform must be specified for 'ota'; add 'platform: esphome' for original OTA functionality.

Fixes: https://github.com/ratgdo/esphome-ratgdo/issues/289

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -6,6 +6,8 @@ external_components:
       url: https://github.com/ratgdo/esphome-ratgdo
     refresh: 1s
 
+safe_mode:
+
 preferences:
   flash_write_interval: 1min
 
--- a/base.yaml
+++ b/base.yaml
@@ -45,6 +45,9 @@ api:
             id($id_prefix).clear_paired_devices(ratgdo::PairedDevice::ACCESSORY);
           }
 
+ota:
+  - platform: esphome
+
 sensor:
   - platform: ratgdo
     id: ${id_prefix}_openings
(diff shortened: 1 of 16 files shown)


bdraco:
This can probably go in base.yaml instead

flavio-fernandes:
> This can probably go in base.yaml instead

Hmm... that would mean that the entire `ota` section should be moved from all the files I changed, right @bdraco ?

Do you agree with that approach, @PaulWieland ?

PaulWieland:
Yes it should be part of the base yamls

flavio-fernandes:
> Yes it should be part of the base yamls

Gotcha!  @bdraco + @PaulWieland :  PTAL and see if I got it right.

bdraco:
Looks correct.

Will have to dig into why the firmware doesn't build. I don't think its specific to this change

bdraco:
https://github.com/ratgdo/esphome-build-action has been updated but looks like there is still an issue with safe_mode

bdraco:
I need to recreate the PR on the main repo

bdraco:
superceeded by https://github.com/ratgdo/esphome-ratgdo/pull/293

Thanks @flavio-fernandes

flavio-fernandes:
> superceeded by #293
> 
> Thanks @flavio-fernandes

NP. Thank you for fixing the build issue!

Issue #146, opened 2023-12-17
Door close command does not work all others do


alrassia:
I have a chamberlain ML700EV with 8 terminals and a ratgdo 2.5i with latest esp software.
My setup features no wall buttons and i previously only used the radio remotes. Type TX2REV with only one button for open and close.
![image](https://github.com/ratgdo/esphome-ratgdo/assets/131307925/a418715d-920c-4327-a42c-f314bf8e26a7)

I got the ratgdo to work with the wiring setup for the 14 terminal setup (use green 0 iso of red terminal for the serial line).

The status is read correctly and commands work, except the close command. Open, stop, toggle,light work ok. Just close results in no action. 

Any clues what is going on here or what might be the cause? Could it maybe use a different close command byte than other models?
Any logs i could run to help provide more insight?

PaulWieland:
Capture the serial log so we can see what’s happening.

alrassia:
I hope this covers the requested logs (VERY_VERBOSE log level), otherwise please let me know what else to do.

[logs_ratgdov25i-f614ca_logs.txt](https://github.com/ratgdo/esphome-ratgdo/files/13732507/logs_ratgdov25i-f614ca_logs.txt)

mariusmuja:
I don't see anything unusual in the logs, other than the fact that your GDO seems to be completely ignoring the close command. 

We might need to add a "quirks" feature flag or something like that for cases like this... such as `USE_TOGGLE_FOR_CLOSE` in this particular case.

P.S. If you feel like experimenting you can see if maybe there's another value for `DOOR_CLOSE` that works for your GDO. You'll need to change the constant here https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.h#L52 and recompile/re-flash to test.

mulcmu:
To experiment quicker you can add this to your yaml under `api:` and then call the service from HA developer tools.
```
  services:
  - service: service_send_door_command
      variables:
        door_data: int
      then:
        - lambda: !lambda |-
            if(door_data >= 0) {
              id($id_prefix).door_command(door_data);  
```

I tried 0x4 to 0xF and 0xFF on my gdo with door open.  No response

alrassia:
Thx I will try to give

> To experiment quicker you can add this to your yaml under `api:` and then call the service from HA developer tools.
> 
> ```
>   services:
>   - service: service_send_door_command
>       variables:
>         door_data: int
>       then:
[...]

PR #298, opened 2024-07-02
Auto detect obstruction sensors

PaulWieland:
When no obstruction sensors are installed, sec+ 2 doors ignore the discrete close command (see #146). The MQTT firmware does not have this problem because it only uses toggle commands. The toggle command is skipped if the door is already in the desired state, which has the same end result as using discrete commands.

Since the toggle command is always accepted by the door, this PR is designed to detect a low pulse on the obstruction sensor input, and if one is found, assume that the sensor is present. If the sensor is present, use the discrete close command, and if not, use the toggle command instead.

It also will report the obstruction status from the status packet if no sensor is detected.

Much testing required before merge.


agoode:
Is there any way to continue to read the obstruction from the status packet even with obstruction sensors connected? I prefer to report obstruction that way so that I have a clear idea of what the door opener thinks. (This was useful when I had a bad sensor that was triggering a door re-open but not being detected with the ratgdo obstruction input.)

In my yaml I currently have this to enable obstruction-from-status:
```yaml
ratgdo:
  input_obst_pin: none
```

PaulWieland:
Not really. The obstruction from status is _really_ slow which is why I designed ratgdo to detect from the hardware sensors directly. It needs to be one or the other or you will get conflicting messages as the status update lags behind the actual sensor state.

agoode:
Seems reasonable. It might be nice to (optionally) expose both obstruction sensors. But I'll assume for now it's one or the other.

components/ratgdo/ratgdo.cpp
```diff
@@ -471,7 +460,12 @@ namespace ratgdo {
             return;
         }
 
-        this->door_action(DoorAction::CLOSE);
+        if(this->obstruction_sensor_detected_){
+			this->door_action(DoorAction::CLOSE);
+        }else if(*this->door_state == DoorState::OPEN){
+			ESP_LOGD(TAG, "No obstruction sensors detected. Close using TOGGLE.");
+            this->door_action(DoorAction::TOGGLE);
+        }
```
bdraco: `clang-format --style=Webkit -i  components/ratgdo/*.h components/ratgdo/*.cpp components/ratgdo/**/*.cpp components/ratgdo/**/*.h  `

bdraco: nevermind, I'll just realized the pre-commit hooks were not working so I fixed that and it will be auto now

PR #265, opened 2024-04-27
Improve compatibility for non-dry-contact configurations.

kpfleming:
* Don't compile any dry-contact code unless PROTOCOL_DRYCONTACT is defined.

* Use BinarySensor instead of GPIOBinarySensor to allow the user to use a template binary sensor (or any other type) if they wish. This is the recommendation of the ESPHome maintainers, and also aligns the C++ code with the Python configuration code (which will accept any type of binary sensor).

Closes #264.

PR #300, opened 2024-07-03
Revert ifdefs added in #265

bdraco:
All boards support dry contacts, the ifdefs are incorrect


kpfleming:
I am using this component in my own configuration where I do not configure dry-contact mode at all. If the intent of this component is to only be used in the configurations provided in this repository, that should be documented.

kpfleming:
Ahh... never mind. My configuration compiles fine with this change in place.

PR #337, opened 2024-11-10
ratgdo32disco

PaulWieland:


components/ratgdo/ratgdo.cpp
```diff
@@ -640,11 +746,14 @@ namespace ratgdo {
     }
     void RATGDOComponent::subscribe_door_state(std::function<void(DoorState, float)>&& f)
     {
+        static int num = 0;
+        auto name = "door_state" + std::to_string(num++);
```
bdraco, line 746: Maybe its the jet lagged, but I'm not following the reason for changing this.

components/ratgdo/sensor/__init__.py
```diff
@@ -38,3 +40,15 @@ async def to_code(config):
     await cg.register_component(var, config)
     cg.add(var.set_ratgdo_sensor_type(config[CONF_TYPE]))
     await register_ratgdo_child(var, config)
+
+    if config['type'] == 'distance':
```
bdraco: Probably needs `black` run on all the `python` code as I'm not sure `pre-commit` does it automatically

components/ratgdo/switch/ratgdo_switch.cpp
```diff
@@ -32,6 +37,9 @@ namespace ratgdo {
             } else {
                 this->parent_->inactivate_learn();
             }
+        } else if (this->switch_type_ == SwitchType::RATGDO_LED) {
+            this->pin_->digital_write(state);
+            this->publish_state(state);
```
bdraco, line 42: I _think_ this is safe here, but need to double check when I'm not in flight

components/ratgdo/ratgdo.cpp
PaulWieland: [Its necessary in order to have multiple subscriptions. ](https://github.com/ratgdo/esphome-ratgdo/commit/c552c5bf9ceb9f79fef318c1e0dd714131e2bac6#comments)

components/ratgdo/switch/ratgdo_switch.cpp
PaulWieland: [Similar approach as dry contact outputs.](https://github.com/ratgdo/esphome-ratgdo/blob/ec3ba31ce8ebaadefbe6855309ed022d45a80c64/components/ratgdo/dry_contact.cpp#L105)

components/ratgdo/sensor/__init__.py
bdraco: I took care of adding it to pre-commit so the CI will take care of it automatically now once this PR is up to date with `main`

PaulWieland:
@bdraco okay with you to merge this PR?

bdraco (approved):

bdraco:
> @bdraco okay with you to merge this PR?

LGTM.  Also quick note that anything with conflicts doesn’t show up in my review queue

PR #338, opened 2024-11-12
ratgdo32 post merge

PaulWieland:
merge this after the ratgdo32 branch is merged into main.

Change deployment branch to `main` from `ratgdo32` in the environment: https://github.com/ratgdo/esphome-ratgdo/settings/environments/1092192999/edit

image


components/ratgdo/sensor/__init__.py
```diff
@@ -5,6 +5,8 @@
 
 CONF_DISTANCE = "distance"
 
+CONF_DISTANCE = "distance"
+
```
bdraco:
```suggestion
```

bdraco (approved):

PaulWieland:
I just reverted the deployment branch to `main`

bdraco:
I recreated `ratgdo32` from `main` and made it a protected branch since it auto deleted when I merged the other PR and there might be users who are still pointing at it

bdraco:
Everything seemed to deploy OK and I updated a few installs. All seems good.

I also added a confirmation before deploy to the env so I could check that everything looked ok before it pushed out

PR #396, opened 2025-04-02
Enhance OPEN state to report position

breakingspell:
This PR: 

- Sets the cover state to OPEN after STOP action at a target position
  - This closes the OPENING state loop present in the drycontact logic (and possibly the other protocols) 
- Reports back the cover position rather than 1.0 value in the OPEN state.
  - Visible in Home Assistant, albeit clamped to a lower value than actually targeted. 
  - Maybe return the target value instead, assuming the door is correctly calibrated+positioned.

image

Fixes:
- https://github.com/ratgdo/esphome-ratgdo/issues/393
- https://github.com/ratgdo/esphome-ratgdo/issues/390

----

However, it also breaks these dry contact limit checks due to the OPEN/UNKNOWN states when partially open. 
I've commented them out in my tests, but it will need addressed before the OPEN state can be modified (maybe not in this PR). Checking the physical limit sensors rather than reported state might be all it needs. 

https://github.com/ratgdo/esphome-ratgdo/blob/94fa006eb650312b6a5f3f491697dc167e6122bb/components/ratgdo/dry_contact.cpp#L93-L100

I'll undraft the PR once I reconcile that, but please let me know if there's any context I'm missing :slightly_smiling_face:


components/ratgdo/ratgdo.cpp
```diff
@@ -626,6 +626,8 @@ namespace ratgdo {
         this->door_action(delta > 0 ? DoorAction::OPEN : DoorAction::CLOSE);
         set_timeout("move_to_position", operation_time, [=] {
             this->door_action(DoorAction::STOP);
+            this->received(DoorState::OPEN);
+            this->cancel_position_sync_callbacks();
```
breakingspell: Pretty sure this is a redundant cancel call, will remove

breakingspell:
Not sure if the latest commit encroaches on other protocols, but I moved the OPEN position override to the STOP/TOGGLE action handler.

breakingspell:
This has been working very well for a week, with two outstanding issues:

- The percentage returned is sometimes behind the door state 
  - Requesting 15% can return 15, 14, 12, 10, etc. The door is actually open in the proper position but the counter is rounded down. 
  - Rounding the output to multiples of 5/10% may present better and correct for this, but masks the actual returned position. This may also require modifying the cover/position calculations, unnecessarily. 

- Bypassing ratgdo and using the wall button or RF remote to stop the door partially will replicate the original issue. 
  - The `Opening` status and counter are triggered by the `Closed` limit sensor turning off, not by the wall button/remote. The other sensor is not encountered, so the status stays in `Opening` and counts up to 97%. Same with `Closing` inversely.
  - Haven't dug in yet, but not sure if this can be accounted for without substantial work regarding dry contact wall buttons, if it's physically possible for ratgdo to monitor button presses. The remotes/paired cars operate entirely within the GDO controller and would be inaccessible to ratgdo..
    - If not, this could be accounted for with a timeout to the [unknown position state](https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.cpp#L180), using a 50% best guess to indicate the door was last opened manually and could be in any position. 
    -  ratgdo's next operation would need to be `Toggle` in order to encounter a sensor and reorient.

azilberbaum:
> Edit: brainstorm on the second issue, we could potentially handle the `Opening/Closing` state timeout based on if it were invoked by external sensor change (limit switch), versus a firmware button/function call. Probably a lot simpler than I had initially thought. This could also be handled in a separate merge.

I think this is a good approach, if ever the limit sensors see the door is open or closed it should update the position.

breakingspell:
> > Edit: brainstorm on the second issue, we could potentially handle the `Opening/Closing` state timeout based on if it were invoked by external sensor change (limit switch), versus a firmware button/function call. Probably a lot simpler than I had initially thought. This could also be handled in a separate merge.
> 
> I think this is a good approach, if ever the limit sensors see the door is open or closed it should update the position.

Yes, subscribing to the Limit Switch's states and handling the external state change as an OPEN/CLOSE would resolve this. 
As it is, that issue is the only one I encounter in day-to-day, housemates use the GDO wall button to partially open instead of scene switch in other room, so the usual position reporting is not handled.

Issue #393, opened 2025-04-01
Endless Opening status - Dry contact wiring


breakingspell:
Hello, this is in addition to https://github.com/ratgdo/esphome-ratgdo/issues/344, and may be considered a duplicate if we include this state in the list.

Positional percentage setting DOES work on latest firmware for an uncommon (Sommer) dry-contact opener. However, once the target position is reached, the esp32 continues to log and report the OPENING/CLOSING operation, with the percentage continuing until hanging. 

The garage DOES stop at the proper position, but the reported status de-syncs. The door will be at the desired "15% Open", while the device and HA integration report it at "97% Opening". 

The hardware logs indicate that while it receives the 15% position command, and calculates the travel time properly: 

```
19:25:45	[D]	[cover:076]	'Door' - Setting
19:25:45	[D]	[cover:084]	  Position: 15%
19:25:45	[D]	[ratgdo:624]	Moving to position 0.15 in 2.2s
19:25:46	[D]	[binary_sensor:036]	'Close limit switch': Sending state OFF
19:25:46	[D]	[ratgdo_dry_contact:053]	Set close_limit_reached to 0
19:25:46	[D]	[ratgdo:113]	Door state=OPENING
19:25:46	[D]	[cover:170]	'Door' - Publishing:
19:25:46	[D]	[cover:173]	  Position: 0%
19:25:46	[D]	[cover:186]	  Current Operation: OPENING
19:25:46	[D]	[cover:170]	'Door' - Publishing:
19:25:46	[D]	[cover:173]	  Position: 4%
19:25:46	[D]	[cover:186]	  Current Operation: OPENING
19:25:47	[D]	[cover:170]	'Door' - Publishing:
19:25:47	[D]	[cover:173]	  Position: 7%
```

It overshoots the target percentage and continues to climb, when the door is actually correctly stopped. It then hangs out in the OPENING status indefinitely, never publishing an OPEN, but it can be closed as expected via controls or automation. 

```
19:25:59	[D]	[cover:173]	  Position: 90%
19:25:59	[D]	[cover:186]	  Current Operation: OPENING
19:25:59	[D]	[cover:170]	'Door' - Publishing:
19:25:59	[D]	[cover:173]	  Position: 94%
19:25:59	[D]	[cover:186]	  Current Operation: OPENING
19:26:00	[D]	[cover:170]	'Door' - Publishing:
19:26:00	[D]	[cover:173]	  Position: 97%
[...]

Issue #344, opened 2024-11-27
Improve door status when using dry contact control protocol


PaulWieland:
When the door is moving and:
- [ ] a new command from ratgdo is sent, the door status should become stopped
- [ ] a low pulse is detected, the door status should become stopped
- [ ] an obstruction is detected, the door status should become opening

PR #398, opened 2025-04-02
Dry Contact Limit sensor improvements

breakingspell:
All dry-contact setups require two limit switches for the open+closed positions. We can utilize these sensors for additional checks and awareness, especially with the OPEN status potentially representing a partially-opened state. 

- The existing checks for the open+closed state in `dry_contact.cpp` are updated to check the sensors before ignoring the open/close command. 
  - This resolves behavior following a reboot with the door positioned partway, the resulting DoorState::UNKNOWN would cause both cover up+down commands to be declined. 
  - It also prevents additional "close" commands while closed from bouncing the door into an inverted state. 

- The limit sensor methods in `ratgdo.cpp` are updated to manually set the door position for the cover once triggered. 
  - This fixes if out-of-sync due to the wall button/obstacle sensor, but requires a "check-in" to a sensor.
    - This could also be fixed or added to by handling the obstruction sensor event to set `Opening` state.
  - The code states this should ideally be done in `drycontact::sync`, for testing i've set them in the main routine. The sensors and door state could potentially be tracked there more efficiently. 

Prerequisite for: 
https://github.com/ratgdo/esphome-ratgdo/pull/396
https://github.com/ratgdo/esphome-ratgdo/pull/401

### DIFF
--- a/components/ratgdo/dry_contact.cpp
+++ b/components/ratgdo/dry_contact.cpp
@@ -92,12 +92,12 @@ namespace ratgdo {
 
         void DryContact::door_action(DoorAction action)
         {
-            if (action == DoorAction::OPEN && this->door_state_ != DoorState::CLOSED) {
-                ESP_LOGW(TAG, "The door is not closed. Ignoring door action: %s", LOG_STR_ARG(DoorAction_to_string(action)));
+            if (action == DoorAction::OPEN && this->limits_.open_limit_reached) {
+                ESP_LOGW(TAG, "The door is already fully open. Ignoring door action: %s", LOG_STR_ARG(DoorAction_to_string(action)));
                 return;
             }
-            if (action == DoorAction::CLOSE && this->door_state_ != DoorState::OPEN) {
-                ESP_LOGW(TAG, "The door is not open. Ignoring door action: %s", LOG_STR_ARG(DoorAction_to_string(action)));
+            if (action == DoorAction::CLOSE && this->limits_.close_limit_reached) {
+                ESP_LOGW(TAG, "The door is already fully closed. Ignoring door action: %s", LOG_STR_ARG(DoorAction_to_string(action)));
                 return;
             }
 
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -920,6 +920,7 @@ namespace ratgdo {
         dry_contact_open_sensor_ = dry_contact_open_sensor;
         dry_contact_open_sensor_->add_on_state_callback([this](bool sensor_value) {
             this->protocol_->set_open_limit(sensor_value);
+            this->door_position = 1.0;
         });
     }
 
@@ -928,6 +929,7 @@ namespace ratgdo {
         dry_contact_close_sensor_ = dry_contact_close_sensor;
         dry_contact_close_sensor_->add_on_state_callback([this](bool sensor_value) {
             this->protocol_->set_close_limit(sensor_value);
+            this->door_position = 0.0;
         });
     }
 


breakingspell:
Hm, sometimes the door can still report CLOSED state. It probably shouldn't report a definite CLOSED unless the limit switch is engaged, need to find out where it is being set.

PaulWieland:
> I hope I'm understanding the purpose of the checks correctly, but let me know!

The original idea is that a discrete open|close command should not cause the door to move in an unexpected direction. (e.g. open command should never cause the door to close and the close command should never cause the door to open). Thats why discrete commands are ignored if the door state is `UNKNOWN` requiring the user to either `Toggle` the door or physically press the wall button.

This is a trade off, but the expectation is that a partially opened door combined with a ratgdo reboot should be a very rare situation.

breakingspell:
> The original idea was that a discrete open|close command should not cause the door to move in an unexpected direction. (e.g. open command should never cause the door to close and the close command should never cause the door to open). 

Hm, I'm not 100% certain, but this doesn't seem to be working as intended in current main:
- Door is fully closed, Closed sensor on, state reports `Closed`.
- In web interface, a discrete command to close cover will "bounce" the door into a fully-open state (I believe the action is converted to `Toggle`). 
- In HA, the same behavior occurs if you grab the shutter slider, and then release at 0% to "cancel"

With this patch the discrete operations are properly declined by the device when at their limit.

PaulWieland:
> In web interface, a discrete command to close cover will "bounce" the door into a fully-open state (I believe the action is converted to Toggle).

A toggle command is used if no obstruction sensors are detected. Most ROW Sec+ 2 openers don't actually have obstruction sensors, and as such they ignore the discrete close command, but accept a toggle command. For that reason the firmware detects if sensors are present, and if not, it issues a toggle instead of a close. 

I have to check but its possible that's causing this behavior in the dry contact firmware. We might have to put a check on the protocol before using toggle in lieu of close when no sensors are attached.

breakingspell:
> A toggle command is used if no obstruction sensors are detected. Most ROW Sec+ 2 openers don't actually have obstruction sensors, and as such they ignore the discrete close command, but accept a toggle command. For that reason the firmware detects if sensors are present, and if not, it issues a toggle instead of a close. 

I noticed that behavior when testing away from the actual motor unit (with no obstruction sensors). [This remap](https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.cpp#L553) does properly resolve when the obstruction sensors are present, the controller processes a discrete `Close`, but still the door still cycles when cover position 0.0 is requested.

If it's being incorrectly called while intended for Sec+ 2 though, then perhaps it should be put behind a protocol check, but I noticed those are sparsely used, if at all in the main `ratgdo.cpp`. There maybe somewhere I'm missing though, thanks for the time to help look! 

For full context, my GDO is a Sommer Synoris Direct Drive 550 (1042v004), single-button. The ratgdo has no trouble interfacing with it :+1:

breakingspell:
Circling back here, the other dry-contact patches I've drafted are based off this core adjustment to the limit sensor reporting, it should be ready to go on it's own. 

----

> I have to check but its possible that's causing this behavior in the dry contact firmware. We might have to put a check on the protocol before using toggle in lieu of close when no sensors are attached.

I double-checked and can confirm that [`CLOSE` is issued properly](https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.cpp#L554), as the obstruction sensors are present. 

It may still be best to make a protocol distinction rather than checking for existence of a common sensor across protocols, but I don't think it's related to this one, as the correct door action is returned. We can fold that into the PR for the [Obstruction sensor improvements](https://github.com/ratgdo/esphome-ratgdo/pull/401). 

Also, I only see the preprocessor [`#ifdef PROTOCOL_DRYCONTACT`](https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.cpp#L509) used once in this manner, and it's stated it's not an ideal case. If it's fine to use that check again, it would work to isolate the behavior to drycontact. 

----

Even with the protocol distinction made for obstruction sensors, these changes to drycontact's limit sensors will:

- Check the limit sensor state rather than the last known door state before declining an impossible action.
  - The door state could be inaccurate (inverted by obstruction cycle, wall button pressed, reboot while partially open). 
  - Prevents the garage door from improperly `Opening` if `DoorAction::CLOSE` is issued while already closed. 
    - It may be specific to my Sommer GDO, but that scenario will open the door. The opener may be making that call internally, so this ends the invalid action before it can reach the GDO. 

- Set the position float value on sensor trigger
  - This works to "check in" and correct the door state if inaccurate. 

All ears for suggestions and feedback!

breakingspell:
Hey @bdraco, I noticed the rebase from main, went ahead and squashed the commits here. Thanks!

bdraco:
Branches and forks now actually test correctly on the CI. You should be able to update this PR now and the CI will actually test it

breakingspell:
Thanks yes! 

It looks like now that it's properly testing the branch, it's failing due to the recent optimizations. Will re-visit, thanks again for fixing the tests

bdraco:
Hi @breakingspell, thank you for this contribution! I wanted to provide some context on the review status.

Since I don't have a dry contact or v1 hardware setup available for testing, I'm currently limited to reviewing PRs that affect v2 hardware or core functionality that I can verify. This means I can't properly test and validate the changes in this PR myself.

I appreciate your patience while we wait for someone with the appropriate hardware setup to review and test these improvements. The changes look well thought out from a code perspective, but hands-on testing with actual dry contact limit switches is essential for this type of functionality.

If anyone in the community with a dry contact setup could help test these changes, that would be greatly appreciated and help move this forward.

Thanks again for your work on improving the limit sensor handling!

breakingspell:
Hello @bdraco! Thanks for the formal intro and all the hard work, it's my pleasure to meet and help out. No urgency for an upstream merge, anyone that wants to build the project with these branches can check them out. We definitely want to make sure I didn't break anything! 

This PR evaluates the physical limit switches instead of the less-accurate door state (which could be partially open) for the dry contact Open/Close commands. This means it would only affect Dry Contact units that use Obstruction Sensors, or ones with discrete Open/Close buttons. I've detailed my GDO model (Sommer) in the thread, it fits that configuration. 

In the discussion we realized it may be due to a Sec+2 component affecting the action behavior, but even with that addressed, I think it's still a good idea to check the physical sensors. This does seem to correct the behavior for me, the door accurately declines an open command if it's already open, and vice versa.

sjafferali:
Hi! I recently installed a ratgdo v32 board using dry contact controls on my LiftMaster LA400 gate opener and noticed some issues with position detection and state tracking. While investigating, I found this PR and the related issues (#393, #344).

> If anyone in the community with a dry contact setup could help test these changes, that would be greatly appreciated and help move this forward.

Gave this a go. Built firmware from the `drycontact-limit-check` branch, flashed it OTA to my device, and ran through a few things to test. Everything worked without issues.

**Hardware:** ratgdo v32 (ESP32), LiftMaster LA400 gate opener, dry contact mode with two normally-open limit switches (open/closed positions)

**What I tested:**

- Full open/close cycle works correctly. State goes CLOSED→OPENING→OPEN and OPEN→CLOSING→CLOSED as expected, position tracks from 0% to 100% and back.
- Sending an Open command while the gate is already fully open is now properly rejected. I can see the `"The door is already fully open. Ignoring door action: OPEN"` log message from the new limit-switch-based guard.
- Sending a Close command while the gate is already fully closed does nothing (gate doesn't move). The `"already fully closed"` log from this PR's guard doesn't show up in this case because `door_close()` in `ratgdo.cpp` silently falls through before reaching the protocol — `obstruction_sensor_detected` is false on my setup and `door_state != OPEN`, so it never calls `door_action()`. The end result is still correct though.
- Position snaps to exactly 100% when the open limit switch triggers and exactly 0% when the close limit switch triggers.
- Opening to a partial position (40%) still has the runaway position issue where the reported status keeps climbing past the actual position. This is expected since that's what #396 is meant to fix.

@bdraco @breakingspell - Hopefully the testing I did is sufficient to move this PR along, but let me know if there is anything else I can do to help!

breakingspell:
@sjafferali Greatly appreciate the testing and results. It certainly helps, I imagine we'll only need a few more positive reports, including dry contact setups with discrete Open/Close buttons. We'll also need the A-OK from Paul and/or bdraco before this can move to main. 

----

> The "already fully closed" log from this PR's guard doesn't show up in this case because door_close() in ratgdo.cpp silently falls through before reaching the protocol — obstruction_sensor_detected is false on my setup and door_state != OPEN, so it never calls door_action().

Just re-tested with my setup and can confirm up close that it's the bug in the [obstruction sensor logic](https://github.com/ratgdo/esphome-ratgdo/blob/856253c6b6790835ead9f2a86ead3ad99de41e58/components/ratgdo/ratgdo.cpp#L647) that @PaulWieland [noted a while back](https://github.com/ratgdo/esphome-ratgdo/pull/398#issuecomment-2776564346): improperly converting the CLOSE to TOGGLE, and as you noticed, silently failing in this case. 

We had the idea of putting this condition behind a protocol check, so assuming `PROTOCOL_SECPLUSV2` is the only intended target for that behavior, i've added the check to the existing PR for Obstruction fixes: https://github.com/ratgdo/esphome-ratgdo/pull/401. I thought to add it here first, but it's more related to the other topic. 

----

We'll certainly still want the enhanced limit switch checks and position updates, looking forward to more feedback and testing. Thanks again for that!

PR #401, opened 2025-04-05
If obstruction sensor is triggered during CLOSING, set door state to OPENING

breakingspell:
Partially addresses https://github.com/ratgdo/esphome-ratgdo/issues/344. Likely need to handle the low-pulse condition differently. 

Curious if this applies only to dry contact protocol, I noticed the wiring diagrams for the other protocols also include these sensors.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -466,6 +466,12 @@ namespace ratgdo {
                     // if the line is high and was last asleep more than 700ms ago, then there is an obstruction present
                     if (current_millis - last_asleep > 700) {
                         this->obstruction_state = ObstructionState::OBSTRUCTED;
+
+                        if (*this->door_state == DoorState::CLOSING) {
+                            this->received(DoorState::OPENING);
+                            // If obstruction sensor is tripped during closing,
+                            // assume the motor has reversed direction.
+                        }
                     }
                 }
             }


breakingspell:
https://github.com/ratgdo/esphome-ratgdo/blob/0336213ffff28fad01932a1c2efa3333b36efb1a/components/ratgdo/ratgdo.cpp#L560-L566

Haven't dug in yet but I think this incorrectly sets `Closed` state after the (obstruction-rolled) cycle's timeout due to `query_status()` being absent from dry_contact protocol. `send_door_state()` seems to serve the same role, I brainstormed just mirroring that response, but based on a physical sensor check. 

Particularly when the Closed limit sensor is present, it might be best to only declare the door securely closed if the sensor is on.

PaulWieland:
Yes this only applies to the dry contact protocol. Security Plus 1/2 gets its status from a serial data connection to the opener. 

The dry contact firmware should depend on both the open and closed limit switches being present for dry contact doors, we do not support or recommend using ratgdo with one or no switches. The `query_status` isn't implemented because there is no need to "query" a dry contact door since the state of the switches is always known. That's not to say that there isn't a bug being introduced by the timeout.

breakingspell:
> The query_status isn't implemented because there is no need to "query" a dry contact door since the state of the switches is always known. That's not to say that there isn't a bug being introduced by the timeout.

Timing out to a `Closed` state without the additional sensor/status check from the dry contact firmware seems the core issue. It actually represents a "worst case" scenario where the door reports closed but it's not :scream: 

In an effort to keep the main `ratgdo.cpp` as protocol-agnostic as possible, would it be best to mimic the response of the Sec 1/2 `query_status()` in `dry_protocol.cpp` to ensure the same status is returned via sensor check from the firmware?

breakingspell:
This has worked as intended over the last month, in conjunction with the dry contact limit switch PR. There are a few places to look at though:

- We discussed changing this [action mapping](https://github.com/ratgdo/esphome-ratgdo/blob/main/components/ratgdo/ratgdo.cpp#L553) to evaluate the protocol instead of the presence of obstruction sensors common to all three protocols.
- This protocol check could also ensure the drycontact firmware doesn't call `query_status()` inappropriately

PaulWieland:
> Should the the low-pulse condition stop the door

No. The code for interpreting obstructions is already set and should not be duplicated elsewhere. Add a subscription to watch for changes to `ObstructionState`.

breakingspell:
I'm definitely misunderstanding the nature of the low-pulse state. Maybe unnecessary for this PR, noted on the subscription pointer though 🙂

components/ratgdo/ratgdo.cpp
```diff
@@ -561,8 +567,11 @@ namespace ratgdo {
             // query state in case we don't get a status message
             set_timeout("door_query_state", (*this->closing_duration + 2) * 1000, [=]() {
                 if (*this->door_state != DoorState::CLOSED && *this->door_state != DoorState::STOPPED) {
-                    this->received(DoorState::CLOSED); // probably missed a status mesage, assume it's closed
-                    this->query_status(); // query in case we're wrong and it's stopped
+                    // In case the door fully rebounded from an obstruction
+                    if (*this->door_state != DoorState::OPEN) {
+                        this->received(DoorState::CLOSED); // probably missed a status mesage, assume it's closed
```
breakingspell: Hm, this `DoorState::CLOSED` does cause problems for the dry contact firmware, since it could be set without the Closed limit switch being active. 

For security, it's probably best to only set `CLOSED` state if the limit sensor agrees

breakingspell: I've put this behind a protocol check rather than a state check

breakingspell:
Added a fix to address https://github.com/ratgdo/esphome-ratgdo/pull/398#issuecomment-2776564346. It may be worth it's own bugfix PR but for now it's more related to these obstruction fixes than the other branches i'm working on. 

Let me know if there are any questions or feedback here! Forgot to mention I've been running these changes daily since implementing them (on a Sommer Synoris 550 1042v004)

components/ratgdo/ratgdo.cpp
```diff
@@ -560,6 +560,13 @@ namespace ratgdo {
                     // if the line is high and was last asleep more than 700ms ago, then there is an obstruction present
                     if (current_millis - last_asleep > 700) {
                         this->obstruction_state = ObstructionState::OBSTRUCTED;
+#ifdef PROTOCOL_DRYCONTACT
```
PaulWieland: I believe you can move this logic to the dry contact protocol logic instead of using an ifdef in the main ratgdo logic. Use a subscriber to watch the obstruction state.

bdraco:
hey, took a look at this. a few things that would need addressing before it can merge

the base is pretty stale, set_timeout uses string based names but main switched to uint32_t scheduler IDs a while back so this will need a rebase

the ifdef PROTOCOL_SECPLUSV2 guard around the obstruction toggle logic doesnt seem right, secplus1 needs that fallback too. might be better to check the protocol traits instead of ifdefs here

also adding protocol specific ifdefs in ratgdo.cpp goes against the direction the codebase has been moving, protocol specific logic should live in the protocol classes where posible

the same query_status timeout issue exists in door_open() for dry contact but only door_close() is fixed here

would be great to get this rebased onto current main and reworked to fit the current architecture, the core fix for dry contact obstruction handlng is definitely needed

breakingspell:
Thanks a ton for the pointers, I'm glad to help! Will see what I can do to clear these up: 

> the base is pretty stale

Fixed! I realized this branch was tracking my fork's main and not the project's. 

> the ifdef PROTOCOL_SECPLUSV2 guard around the obstruction toggle logic doesnt seem right, secplus1 needs that fallback too. might be better to check the protocol traits instead of ifdefs here

> also adding protocol specific ifdefs in ratgdo.cpp goes against the direction the codebase has been moving, protocol specific logic should live in the protocol classes where posible

I'd assumed Secplus2 not having obstruction sensors was unique to that system, but that was only going by [Paul's comment](https://github.com/ratgdo/esphome-ratgdo/pull/398#issuecomment-2776564346). I'm not sure if it covers both protocols or not, that's definitely a question for those more versed. 

Even so, yes it would definitely be best to move these specific behaviors into the protocol classes and out of the main cpp. I can take a crack at it! 

> the same query_status timeout issue exists in door_open() for dry contact but only door_close() is fixed here

Good catch yes! This should be easy to drop-in once the logic is reworked into classes.  

----

EDIT: Took a quick few tests tonight, created a new protocol method for all three and was able to move the logic into each respective protocol class. Will look into defining this as a trait and setting the behavior behind that.

PR #428, opened 2025-06-23
Reduce memory needed

bdraco:
## Summary

This PR implements comprehensive memory optimizations for the RATGDO component, reducing RAM usage by approximately **101+ bytes** per instance through strategic data structure improvements and protocol-aligned type corrections.

### 1. Enum Size Optimization
- Changed `SensorType` and `RATGDOSensorType` enums to use `uint8_t` base type instead of default `int`
- Changed `CommandType` enum in `secplus1.h` to use `uint8_t` 
- **Savings**: 3 bytes per enum instance

### 2. Dynamic to Static Container Conversion
- Replaced `std::vector<int16_t>` with `std::array<int16_t, 30>` for `distance_measurement`
- Eliminates vector overhead (capacity, size, pointer)
- **Savings**: ~24 bytes

### 3. Boolean Flag Packing
Converted multiple boolean members to bitfields across all protocol classes:

- **DryContact**: 4 bools → 1 byte bitfield (saved 3 bytes)
- **Secplus1**: 3 bools → 1 byte bitfield (saved 2 bytes)  
- **Secplus2**: 1 bool → 1 byte bitfield (future-proof)
- **RATGDOComponent**: 2 bools → 1 byte bitfield (saved 1 byte)

### 4. Class Member Reordering
Reordered class members to minimize padding on 32-bit systems:

- **Secplus1**: Grouped all single-byte members together (saved ~27 bytes)
- **DryContact**: Optimized member layout (saved 4 bytes)
- **Secplus2**: Improved alignment (saved 5 bytes)
- **RATGDOComponent**: Grouped pointers and similar-sized members (saved 2 bytes)

### 5. Flash Memory Optimization
- Added `PROGMEM` attribute to `secplus1_states[]` constant array
- **Savings**: 19 bytes moved from RAM to flash

### 6. Protocol-Aligned Type Corrections
Changed paired device counters from `uint16_t` to `uint8_t`:
- `paired_total`, `paired_remotes`, `paired_keypads`, `paired_wall_controls`, `paired_accessories`
- The garage door protocol only transmits device counts as a single byte (0-255)
- This change both saves memory and correctly matches the protocol specification
- **Savings**: 5 bytes (5 variables × 1 byte each)

## Memory Savings Breakdown

| Component | Optimization | RAM Saved |
|-----------|-------------|-----------|
| Enums | uint8_t base type | ~6 bytes |
| distance_measurement | vector → array | ~24 bytes |
| Boolean packing | bitfields | ~9 bytes |
| Member reordering | alignment | ~38 bytes |
| PROGMEM constants | flash storage | 19 bytes |
| Paired device counters | uint16_t → uint8_t | 5 bytes |
| **Total** | | **~101+ bytes** |

## Impact

For a typical RATGDO installation:
- ESP8266 (80KB RAM): ~0.12% more available RAM
- ESP32 (320KB RAM): Improved memory efficiency
- Reduced memory fragmentation
- Better cache locality due to improved alignment

## Testing

All changes maintain API compatibility and require no configuration changes. The optimizations are transparent to end users while providing better resource utilization on memory-constrained devices.

## Technical Notes

- All type changes maintain protocol compatibility
- The paired device counter change aligns with the actual garage door protocol, which uses single-byte values
- Memory layout optimizations consider 32-bit ESP32/ESP8266 architecture alignment requirements

## Future Optimization Opportunities

Additional optimizations identified but not implemented in this PR:
- Observable pattern optimization (could save ~300-1600 bytes)
- Further PROGMEM usage for string literals
- Reduced `distance_measurement` array size if 30 samples aren't needed
- Time variable optimization (uint32_t → uint16_t for relative timestamps)

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -3,7 +3,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: ram
+      ref: main
     refresh: 1s
   # - source:
   #     type: local
--- a/base.yaml
+++ b/base.yaml
@@ -3,7 +3,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
     refresh: 1s
   # - source:
   #     type: local
--- a/base_drycontact.yaml
+++ b/base_drycontact.yaml
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
     refresh: 1s
   # - source:
   #     type: local
--- a/base_drycontact.yaml
+++ b/base_drycontact.yaml
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: ram
+      ref: main
     refresh: 1s
   # - source:
   #     type: local
--- a/base_secplusv1.yaml
+++ b/base_secplusv1.yaml
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
     refresh: 1s
   # - source:
   #     type: local
--- a/base_secplusv1.yaml
+++ b/base_secplusv1.yaml
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: ram
+      ref: main
     refresh: 1s
   # - source:
   #     type: local
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     -   id: end-of-file-fixer
     -   id: check-added-large-files
 - repo: https://github.com/pre-commit/mirrors-clang-format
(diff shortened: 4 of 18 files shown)


bdraco:
I don't use the distance measurement so this will need testing for that

PaulWieland:
I will update all of my units now. Array of 30 is the minimum, but even then it can be possible to get false readings. One possibility might be to slow down the sampling rate of the distance sensor data.

base.yaml
```diff
@@ -3,7 +3,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
```
```suggestion
      ref: main
```

Revert before merging

PaulWieland (changes requested):
Revert before merging

base_drycontact.yaml
```diff
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
```
bdraco:
```suggestion
      ref: main
```

base_secplusv1.yaml
```diff
@@ -4,7 +4,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: ram
```
bdraco:
```suggestion
      ref: main
```

PR #458, opened 2025-07-28
Reduce memory usage to improve ESP8266 stability

bdraco:
## Problem
ESP8266 devices are experiencing instability and crashes due to RAM exhaustion. The current `observable` implementation uses significant memory even for single-observer cases.

## Summary
This PR introduces a memory-optimized `single_observable` class that reduces RAM usage by ~75% for observables that only need a single subscriber. The original `observable` class uses `std::vector<std::function<void(T)>>` which has significant overhead even for single observers.

## Memory Savings
- **Before**: Each observable uses ~12 bytes (vector on ESP8266) + ~16-24 bytes per observer = 28-36+ bytes
- **After**: Each single_observable uses sizeof(T) + ~16-24 bytes (std::function) = ~20-28 bytes
- **Measured savings**: 376 bytes of free RAM on ESP8266
- std::function still has overhead for storing lambdas with captures that we can't git rid of right now. But this is still a good step forward

## Changes
1. Added `single_observable` class to `observable.h` that stores only one observer
2. Converted 23 observables in `ratgdo.h` from `observable` to `single_observable`
3. Converted `rolling_code_counter_` in `secplus2.h` to use `single_observable`
4. Added ESP_LOGE warning when attempting to subscribe multiple times to a `single_observable`
5. Added Python validation to prevent multiple instances of the same sensor/binary_sensor type
6. Added validation to ensure only one light and one lock component per device

## Safety
- The `single_observable` class logs an error if a second subscriber attempts to subscribe
- Python validation prevents configuration errors that would lead to multiple subscribers
- Kept `observable` for `door_state` and `door_position` which may need multiple observers

## Testing
- Existing functionality preserved - single observer pattern is backward compatible
- Error logging helps identify any missed multi-subscriber cases
- Python validation provides clear error messages at config time

## Performance Benefits
- Reduced memory fragmentation from vector allocations
- Better cache locality (no pointer indirection)
- Faster notification dispatch (direct function call vs vector iteration)

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -6,7 +6,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: single_observable
     refresh: 1s
   # - source:
   #     type: local
--- a/base.yaml
+++ b/base.yaml
@@ -6,7 +6,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: single_observable
+      ref: main
     refresh: 1s
   # - source:
   #     type: local
--- a/components/ratgdo/binary_sensor/__init__.py
+++ b/components/ratgdo/binary_sensor/__init__.py
@@ -7,6 +7,9 @@ from .. import RATGDO_CLIENT_SCHMEA, ratgdo_ns, register_ratgdo_child
 
 DEPENDENCIES = ["ratgdo"]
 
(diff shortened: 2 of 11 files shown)


base.yaml
```diff
@@ -6,7 +6,7 @@ external_components:
   - source:
       type: git
       url: https://github.com/ratgdo/esphome-ratgdo
-      ref: main
+      ref: single_observable
```
```suggestion
      ref: main
```

PR #469, opened 2025-08-23
Simplify configuration by auto-detecting ratgdo_id

kpfleming:
In configurations with only one instance of the 'ratgdo' component (nearly all of them), there is no need for the entity configurations to specify the ID of the component, as the schema can auto-detect it.


bdraco (approved):
Thanks @kpfleming

PR #471, opened 2025-08-29
Implement HomeKit (HAP) support with native garage door opener for ESP32 RATGDO boards

donavanbecker:


bdraco:
Tests are failing

components/homekit/HAPAccessory.cpp
```diff
@@ -0,0 +1,111 @@
+#include "HAPAccessory.h"
```
bdraco: I'm hoping this is just for testing and dev as we can't maintain a homekit implementation. This should use the external component

donavanbecker: @bdraco we are going to have to wait till this is merged: https://github.com/rednblkx/HAP-ESPHome/pull/47

Before we can merge this as this will not build until then.

base_hap.yaml
```diff
@@ -0,0 +1,103 @@
...
+# Setup ID: RGDO (4-character identifier for HomeKit)
+homekit_base:
+  meta:
+    name: "RATGDO HomeKit Bridge"
+    manufacturer: "RATGDO"
+    model: "ESPHome Garage Door Controller"
+    serial_number: "${id_prefix}"
+    fw_rev: "2.0-hap"
+  setup_code: '159-35-728'  # Consider changing for production deployment
+  setup_id: "RGDO"
```
dkerr64, line 47: A hard coded setup ID can cause problems if two brand new devices are powered onto the same network and HomeKit tries to pair with one of them... the other could reply.  I solved this in my HomeKit firmware by generating a unique ID based on the lower 24-bits of MAC address.  The setup ID can be base-62 encoded [A-Z][a-z][0-9] which will encode 22-bits... good enough (I just drop the most significant character of the base-62 encoding.

With that you can then generate a QR code payload (text string X-HM://xxxxx) which you can send to the browser which can then generate a pairing QR code with a javascript library... so adding to Apple Home can be done by scanning the QR code.

PR #482, opened 2025-09-14
Move some more log strings to flash on esp8266

bdraco:


components/ratgdo/secplus2.cpp
```diff
@@ -222,7 +222,7 @@ namespace ratgdo {
             if (kind == PairedDevice::UNKNOWN) {
                 return;
             }
-            ESP_LOGW(TAG, "Clear paired devices of type: %s", PairedDevice_to_string(kind));
+            ESP_LOGW(TAG, "Clear paired devices of type: %s", LOG_STR_ARG(PairedDevice_to_string(kind)));
```
bdraco, line 225: This is a bugfix, it would crash on esp8266 before this fix

PR #496, opened 2025-10-22
improve vehicle detection/bouncing

PaulWieland:
increase the array size to have more samples and switch vehicle detected = true from all_in_range to any_in_range logic.

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -384,11 +384,7 @@ namespace ratgdo {
 
 #ifdef RATGDO_USE_VEHICLE_SENSORS
         this->in_range <<= 1;
-        if (distance <= *this->target_distance_measurement) {
-            this->in_range.set(0, true);
-        } else {
-            this->in_range.set(0, false);
-        }
+        this->in_range.set(0, distance <= *this->target_distance_measurement);
         this->calculate_presence();
 #endif
     }


PaulWieland:
So far its more accurate, but it consumes more RAM because the distance_measurement array was increased from 30 to 120.

PaulWieland:
A smarter way to do this might be to use a bitset instead of an array int16_t. check if the value is in range or not, save in the bitset and then check if any/all/none of the bitset is true or false.

PaulWieland:
bitset approach implemented

components/ratgdo/ratgdo.h
```diff
@@ -78,8 +74,8 @@ namespace ratgdo {
 
 #ifdef RATGDO_USE_DISTANCE_SENSOR
         single_observable<int16_t> target_distance_measurement { -1 };
-        std::array<int16_t, 30> distance_measurement {}; // the length of this array determines how many in-range readings are required for presence detection to change states
-        single_observable<int16_t> last_distance_measurement { 0 };
+        in_range std::bitset<2056> // the length of this bitset determines how many out of range readings are required for presence detection to change states
```
bdraco: What's 2056 intended here. That's ~257 bytes

bdraco: Was 2056 or 256 intended here. 2056 is ~257 bytes

PaulWieland: typo, fixed

PaulWieland:
@dkerr64 FYI, on this change. I am testing with a new set of surfaces and lighting conditions. If implemented, we should apply the same logic to the HK firmware.

dkerr64:
@PaulWieland so you are increasing the number of distance samples to 256, and if any one of those samples falls under the target distance for vehicle presence then it declares that a vehicle is present?

In addition, for purpose of tracking, you are writing to log the percentage of samples "in-range" whenever that percentage changes.

I would assume that when no vehicle is present then (pointing at garage floor) the percentage would be stable at zero.  When a vehicle arrives there will be a flurry of log messages.  Then (hopefully) it should stabilize again at 100%.  But to allow for reflections, etc. it is not required to be constant 100%... anything greater than zero means vehicle present.

dkerr64:
How do we handle situations where there is no vehicle but someone is walking around in the garage and walks under the sensor?

components/ratgdo/ratgdo.cpp
```diff
@@ -381,13 +381,13 @@ namespace ratgdo {
...
-            this->distance_measurement[i] = this->distance_measurement[i - 1];
-        }
-        this->distance_measurement[0] = distance;
 #ifdef RATGDO_USE_VEHICLE_SENSORS
+        this->in_range <<= 1;
+        if (distance <= *this->target_distance_measurement) {
+            this->in_range.set(0, true);
+        } else {
+            this->in_range.set(0, false);
+        }
```
bdraco:
```suggestion
        this->in_range.set(0, distance <= *this->target_distance_measurement);
```

components/ratgdo/ratgdo.cpp
```diff
@@ -396,31 +392,17 @@ namespace ratgdo {
...
-            }
-
-            if (value < *this->target_distance_measurement && value != -1) {
-                all_out_of_range = false;
-            }
-        }
-#endif
-
-        if (all_in_range)
+        if (this->in_range.any())
```
bdraco: Is there a risk of false positives?

Completely untested but maybe something like:

```diff
diff --git a/components/ratgdo/ratgdo.cpp b/components/ratgdo/ratgdo.cpp
index 97ff868..2db41c4 100644
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -40,6 +40,7 @@ namespace ratgdo {
 #ifdef RATGDO_USE_VEHICLE_SENSORS
     static const int CLEAR_PRESENCE = 60000; // how long to keep arriving/leaving active
     static const int PRESENCE_DETECT_WINDOW = 300000; // how long to calculate presence after door state change
+    static constexpr float PRESENCE_DETECTION_THRESHOLD = 5.0f; // minimum percentage of samples required to detect vehicle
 #endif
 
     void RATGDOComponent::setup()
@@ -392,11 +393,14 @@ namespace ratgdo {
 #ifdef RATGDO_USE_VEHICLE_SENSORS
     void RATGDOComponent::calculate_presence()
     {
-        if (this->in_range.any())
+        float percent = (float)this->in_range.count() / this->in_range.size() * 100.0f;
+
+        // Use threshold to reduce false positives while still being responsive
+        if (percent >= PRESENCE_DETECTION_THRESHOLD)
             this->vehicle_detected_state = VehicleDetectedState::YES;
-        if (this->in_range.none())
+        if (percent == 0.0f)
             this->vehicle_detected_state = VehicleDetectedState::NO;
-        float percent = (float)this->in_range.count() / this->in_range.size() * 100.0f;
+
         static float last_percent = -1.0f;
         if (percent != last_percent) {
             ESP_LOGD(TAG, "pct_in_range: %f", percent);
```

PaulWieland:
> @PaulWieland so you are increasing the number of distance samples to 256, and if any one of those samples falls under the target distance for vehicle presence then it declares that a vehicle is present?

Yes. There are situations where ambient light creates a ton of noise. The before logic required all samples to be in range for the vehicle to be present, and then all samples to be out of range for it to be not present.

> In addition, for purpose of tracking, you are writing to log the percentage of samples "in-range" whenever that percentage changes.

This is for debugging purposes, and might not remain active. I have a test surface setup now that consistently reports 80 - 85% when under the sensor.
 
> I would assume that when no vehicle is present then (pointing at garage floor) the percentage would be stable at zero. When a vehicle arrives there will be a flurry of log messages. Then (hopefully) it should stabilize again at 100%. But to allow for reflections, etc. it is not required to be constant 100%... anything greater than zero means vehicle present.

Exactly.

PaulWieland:
> How do we handle situations where there is no vehicle but someone is walking around in the garage and walks under the sensor?

This would briefly trigger a false positive, but automations should be tied to the arriving and leaving sensors, not the detected sensor. The arriving and leaving sensors do not fire unless the door was activated within the timeout window.

components/ratgdo/ratgdo.cpp
PaulWieland: The `PRESENCE_DETECTION_THRESHOLD` idea is exactly where I planned to go next if I end up getting false positives in my test. In addition I was contemplating adding a slider range user input to control the % threshold, so users could tune it for their situation, but I'm not sure if its necessary yet. I want to see what kind of %s the debug outputs during the next 24-48 hours as the sun comes out in the "worst case scenario" that I have staged for testing right now.

dkerr64:
> > How do we handle situations where there is no vehicle but someone is walking around in the garage and walks under the sensor?
> 
> This would briefly trigger a false positive, but automations should be tied to the arriving and leaving sensors, not the detected sensor. The arriving and leaving sensors do not fire unless the door was activated within the timeout window.

Yes, you are right, we do also factor in door being activated.

components/ratgdo/ratgdo.cpp
```diff
@@ -396,31 +392,17 @@ namespace ratgdo {
...
             this->vehicle_detected_state = VehicleDetectedState::YES;
-        if (all_out_of_range)
+        if (this->in_range.none())
             this->vehicle_detected_state = VehicleDetectedState::NO;
-
-        // auto k = this->distance_measurement;
-        // ESP_LOGD(TAG,"measure: %i,%i,%i,%i,%i,%i,%i,%i,%i,%i; target: %i; all_in: %s; all_out: %s;", k[0],k[1],k[2],k[3],k[4],k[5],k[6],k[7],k[8],k[9], *this->target_distance_measurement, all_in_range ? "y" : "n", all_out_of_range ? "y" : "n");
+        float percent = (float)this->in_range.count() / this->in_range.size() * 100.0f;
```
dkerr64: integer arithmetic  would be more efficient?
```
int percent = this->in-range.count() * 100 / this->in-range.size();
```
Assumes that you don't care about fractions of percent.

dkerr64: Also... on first boot we get 100 messages (256 if using `float`) as bits are added to the set.  Maybe we keep a count and divide by
```
std::min(count, this->in-range.size());
```
Can a `std::bitset` be resized?  Not sure how good compiler optimization is, but may be more efficient to
```
#define SET_SIZE 256
std::min(count, SET_SIZE);
```
This of course is only of value during initial startup while we sample the first 256 data points.

PaulWieland: > Can a std::bitset be resized 

No. It would have to be vector instead I think.

PaulWieland:
This is an imperfect chart, because it only plots when the value changes. However I can see that it drops to 20% at 9am, which is when the sun shines through the window onto my test rig. The test rig is a black, highly reflective, angled piece of metal on a stand that receives natural light through a large window.

The 5% threshold might be good enough to ignore people walking under the sensor while still providing quick enough vehicle detection.

image

components/ratgdo/ratgdo.cpp
PaulWieland: I Implemented this suggestion. I was getting false positives on my control test which is pointed at the floor. Apparently a floating spec of dust was enough to trigger a reading.

bdraco:
On a side note https://github.com/ratgdo/esphome-ratgdo/pull/487 should hopefully be mergeable with ESPHome 2025.11.x. I reworked one of the filters that was consuming 25000 bytes of RAM for the 5000 samples

PaulWieland:
I'm going to merge this. So far it shows no downside compare to the previous implementation and challenging environments are definitely improved.

dkerr64:
> I'm going to merge this. So far it shows no downside compare to the previous implementation and challenging environments are definitely improved.

Okay, I'll work on HomeKit version.

components/ratgdo/ratgdo.cpp
```diff
@@ -396,31 +393,18 @@ namespace ratgdo {
...
-#endif
+        int percent = this->in_range.count() * 100 / this->in_range.size();
+        static int last_percent = -1;
 
-        if (all_in_range)
+        if (percent >= PRESENCE_DETECTION_THRESHOLD)
             this->vehicle_detected_state = VehicleDetectedState::YES;
-        if (all_out_of_range)
+        if (percent == 0)
             this->vehicle_detected_state = VehicleDetectedState::NO;
```
dkerr64, line 402: By this logic if percent is between zero and threshold then the detected state is unchanged... remains in previous state.  Thus.

- Transition from NO to YES requires >= threshold
- Transition from YES to NO requires zero percent (ie, not < threshold)

Is that the intended design?

PaulWieland: yes.

When everything in the bitset is out of the threshold range (or invalid), then the vehicle detected state is no. When more than 5% of the bitset is in range then the vehicle detected state is yes.

PR #498, opened 2025-10-24
single_observable -> observable

PaulWieland:
Any `RATGDOComponent::subscribe_*` which uses a `static int num = 0;` is assumed to have multiple subscribers


PaulWieland:
This reverts some properties to `observable` that were switched to `single_observable` in #458

bdraco (approved):

PR #501, opened 2025-10-27
Improve debounce

PaulWieland:
adds a separate PRESENCE_DETECTION_OFF_THRESHOLD which requires the entire bitset to be false for X iterations before switching VehicleDetectedState

@dkerr64 FYI


dkerr64:
Thanks for tagging me.  I'll take a look and include in HomeKit version

Issue #425, opened 2025-06-12
Garage Door randomly locked


nickp85:
I have Ratgdo v2.5i hooked up to my Liftmaster that's about 5 years old. I don't usually keep ESPHome running on my network unless I specifically need to do somethin so Ratgdo just sits there doing it's thing.

The other day I randomly got an alert from the Home app on my phone that the garage door had been locked. I don't have any configured automations to do that and frankly I didn't even know it was possible. I am the only one with access to the home and the ratgdo is connected to a secure wifi.

I didn't even realize what this locked message meant so I ignored it. Today I found that all my remotes were locked and only the wall mounted remote would work and it had a flashing light. No one had engaged the lock, I live alone.

When I booted up ESPHome sure enough I could see a lock engaged. I disabled it and am fine again.

How would the lock engage by itself? Is this a bug somewhere? I've been using Ratgdo for users and have never seen this happen.

Thanks

mwaterbu:
I have seen this happen after a power outage. Is it possible that the power went out?

KnightRdr23:
There was no power outage

Thanks

mullender:
I have been experiencing the same issue, and haven't been able to figure out what triggers the lock remotes action.

For now I have an automation in home assistant that will unlock the remotes as soon as they get locked. This prevents the frustration when the remote opener button doesn't work when you pull up.

cptsk1ttles:
for me this happens anytime I restart homeassitant for an update.

mullender:
Is there a way we can completely disable this function in the Ratgdo? or add some more logging on what could possibly trigger this action?

mathman991177:
I am having the same experience using home assistant and chamberlain opener. 

The automation I added of unlocking when state is locked is working but I don’t like the thought of randomly locking out all my remotes

PaulWieland:
I would suggest running a long term log capture to determine the packet which is triggering the remote to lock. Unfortunately I do not have a power shell script for Windows yet, however if you have a mac or linux computer you can use [this ratgdo logger script](https://github.com/ratgdo/remote-logging)

mullender:
Thanks for the log capture instructions.
I have started the logger, and now we wait..

brianf:
I'm having this issue as well with a 8500WGD and the Ratgdo Disco

brianf:
[...]

PR #503, opened 2025-10-31
Update ratgdo_lock.cpp

PaulWieland:
This PR is intended to fix issue #425 

Previously ratgdo would receive a lock status update from the GDO and then send the same exact lock status command back to the GDO, which is unnecessary.

On top of that, if a bad packet is sent which is interpreted as the door lock status being flipped, ratgdo would actually command the door to change it's lock status.


bdraco (approved):

bdraco:
Looks like a good fix. control seems completely wrong. It should just publish the state not loop it back to the device

PR #511, opened 2025-11-29
Fix ESP32 v2.5 obstruction logic; add obst_sleep_low config

jbdman:
### Fix ESP32 obstruction logic on v2.5 boards; add `obst_sleep_low` config option

Some v2.5 boards using ESP32 D1 Mini controllers were reporting spurious obstruction events due to the ESP32 "asleep" logic being inverted from earlier versions. This behavior appears to have been introduced in commit **[f68d126](https://github.com/ratgdo/esphome-ratgdo/commit/f68d126)** (ratgdo32 work).

This PR adds a runtime boolean config option:

```yaml
ratgdo:
  obst_sleep_low: true
```

When `true`, the obstruction sensor treats LOW as the "asleep" state (original v2.5 behavior). The default is platform-aware via `cv.SplitDefault`:

- **ESP32**: `false` (HIGH = asleep) — preserves current behavior
- **ESP8266**: `true` (LOW = asleep) — preserves current behavior

The `v25board_esp32_d1_mini.yaml` board config sets `obst_sleep_low: true` to fix the false obstruction issue on those boards.

This option is intentionally **not** exposed as a substitution in `base.yaml`. Using a substitution would require a single default value, bypassing `SplitDefault` and breaking the per-platform default for one platform. Instead, boards that need to override the default add a `ratgdo:` block with a matching `id`. Users can do the same in their own config:

```yaml
# Example: override in user config for an ESP32 v2.5 board
ratgdo:
  id: ratgdov25
  obst_sleep_low: true
```

#### Changes

- Replace compile-time `#ifdef USE_ESP32` with a runtime check against `flags_.obst_sleep_low`
- Add `obst_sleep_low` boolean config option with `cv.SplitDefault` for per-platform defaults
- Add `set_obst_sleep_low()` setter and bitfield flag to `RATGDOComponent`
- Set `obst_sleep_low: true` in `v25board_esp32_d1_mini.yaml` via a `ratgdo:` block with matching `id`

Tested on two ESP32 v2.5 boards where this resolves the false obstruction issue.

### DIFF
--- a/base.yaml
+++ b/base.yaml
@@ -27,6 +27,7 @@ ratgdo:
   input_gdo_pin: ${uart_rx_pin}
   output_gdo_pin: ${uart_tx_pin}
   input_obst_pin: ${input_obst_pin}
+  obst_sleep_level: ${obst_sleep_level}
   on_sync_failed:
     then:
       - homeassistant.service:
--- a/base.yaml
+++ b/base.yaml
@@ -1,4 +1,8 @@
 ---
+substitutions:
+  # Default Value(s)
+  obst_sleep_level: high
+
 esphome:
   min_version: 2025.7.0
 
--- a/base.yaml
+++ b/base.yaml
@@ -1,7 +1,7 @@
 ---
 substitutions:
   # Default Value(s)
-  obst_sleep_level: high
+  obst_sleep_low: false
 
 esphome:
   min_version: 2025.7.0
@@ -31,7 +31,7 @@ ratgdo:
   input_gdo_pin: ${uart_rx_pin}
   output_gdo_pin: ${uart_tx_pin}
   input_obst_pin: ${input_obst_pin}
-  obst_sleep_level: ${obst_sleep_level}
+  obst_sleep_low: ${obst_sleep_low}
   on_sync_failed:
     then:
       - homeassistant.service:
--- a/components/ratgdo/__init__.py
+++ b/components/ratgdo/__init__.py
@@ -84,9 +84,7 @@ CONFIG_SCHEMA = cv.All(
                 cv.none, pins.gpio_input_pin_schema
             ),
             cv.Optional(CONF_OBST_SLEEP_LVL, default=DEFAULT_OBST_SLEEP_LVL): cv.one_of(
-                OBST_SLEEP_LEVEL_LOW,
-                OBST_SLEEP_LEVEL_HIGH,
-                lower=True
+                OBST_SLEEP_LEVEL_LOW, OBST_SLEEP_LEVEL_HIGH, lower=True
             ),
             cv.Optional(CONF_DISCRETE_OPEN_PIN): pins.gpio_output_pin_schema,
             cv.Optional(CONF_DISCRETE_CLOSE_PIN): pins.gpio_output_pin_schema,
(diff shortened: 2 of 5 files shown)


bdraco (changes requested):
Please make this a YAML configuration option in the schema and set it for the esp32 2.5 configs

jbdman:
@bdraco Let me know if this matches what you were thinking. I implemented it as a compile-time option, since that seemed like the safest and most consistent approach here.

I’ve only applied and tested this on the specific ESP32 v2.5 board I have access to, so I’m absolutely open to adjustments or improvements.

bdraco (changes requested):
A few issues to fix:

1. **Substitution name mismatch** - `v25board_esp32_d1_mini.yaml` uses `obst_sleep_lvl` but `base.yaml` expects `obst_sleep_level`. The override won't work due to this typo.

2. **Use `cg.add_define` instead of `cg.add_build_flag`** - This is the ESPHome idiomatic approach:
   ```python
   cg.add_define("RATGDO_OBST_SLEEP_LOW")
   ```

3. **Simplify the config check** - Since the config has a default value, `config.get()` will always return a value. Can simplify to:
   ```python
   if config[CONF_OBST_SLEEP_LVL] == OBST_SLEEP_LEVEL_LOW:
       cg.add_define("RATGDO_OBST_SLEEP_LOW")
   ```

bdraco (approved):

bdraco:
@jbdman I cleaned up the PR
Thanks for bringing this

kaechele:
Possibly dumb question:

Wouldn't https://github.com/ratgdo/esphome-ratgdo/blob/77db677ec0572fcb96fa6b423279f59d6ca8ff41/components/ratgdo/ratgdo.cpp#L86-L88 prevent the obstruction pin from ever going low when using ESP32 on boards that don't have the inverted logic (i.e. extra MOSFET on the obstruction pin) like the v2.5?

If that is the case, shouldn't this rather be handled via the GPIO Pin setup and using `inverted: True` and a conditional `mode: INPUT` / `mode: INPUT_PULLUP`?

Setting up the pin correctly would then also automatically make https://github.com/ratgdo/esphome-ratgdo/blob/77db677ec0572fcb96fa6b423279f59d6ca8ff41/components/ratgdo/ratgdo.cpp#L91-L93  trigger on the desired edge as `attach_interrupt` appropriately considers inverted state from the pin setup.

EDIT: I'm not an Electrical Engineer, but it just occurred to me that with the original design of two 10kΩ resistors as a voltage divider for the obstruction signal we effectively create a situation in which the nominally ~45kΩ pull-up resistor in the ESP32 will raise the floating voltage of a "low" signal from the obstruction sensor, potentially causing misreads.

PaulWieland:
All ratgdo32 boards use inverted logic.

kaechele:
That's what I thought. I was mainly referring to the following configurations:
- [v25board_esp32_d1_mini.yaml](https://github.com/ratgdo/esphome-ratgdo/blob/main/static/v25board_esp32_d1_mini.yaml)
- [v2board_esp32_d1_mini.yaml](https://github.com/ratgdo/esphome-ratgdo/blob/main/static/v2board_esp32_d1_mini.yaml)
- [v2board_esp32_lolin_s2_mini.yaml](https://github.com/ratgdo/esphome-ratgdo/blob/main/static/v2board_esp32_lolin_s2_mini.yaml)

I believe those would be configurations where the internal pull-up being enabled could cause issues.

PR #516, opened 2025-12-10
check parity of decoded wireline messages

PaulWieland:


components/ratgdo/secplus2.cpp
```diff
@@ -348,7 +348,11 @@ namespace ratgdo {
             uint64_t fixed = 0;
             uint32_t data = 0;
 
-            decode_wireline(packet, &rolling, &fixed, &data);
+            int err = decode_wireline(packet, &rolling, &fixed, &data);
+            if (err < 0) {
+                ESP_LOGW(TAG, "Decode failed (parity error or invalid frame)");
+                return {};
```
PaulWieland: @bdraco do you see anything wrong with this change? I'm primarily worried about returning `{}` having some unintended consequence when a bad packet is received.

PR #602, opened 2026-04-12
If Sec+2 door has an internal time to close delay

PaulWieland:
Some openers have a built in delay when they receive discrete close commands. They wait a pre-set amount of time before the door physically starts closing. When this happens, the door reports it is closing, waits, and then turns the motor on.

This delay throws off the reported % position of the door.

This fix attempts to base the door close start time off of when the motor state turns to ON.

### DIFF
--- a/components/ratgdo/observable.h
+++ b/components/ratgdo/observable.h
@@ -152,6 +152,8 @@ public:
         this->observer_ = Callback<T>::create(std::forward<F>(observer));
     }
 
+    bool has_observer() const { return static_cast<bool>(this->observer_); }
+
     void notify() const
     {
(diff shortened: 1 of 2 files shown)


PaulWieland:
Testing still required

bdraco:
This likely is a breaking change because everybody's door that doesn't have this problem is calibrated with the wrong time after this.

We may need to do a one time clear of the timings by versioning the preference

PaulWieland:
I have to do more testing, but from what I recall the majority of sec + 2 openers report motor on within 1 second or less of door state changing to opening. I’m hoping the impact is minimal.

Issue #179, opened 2024-01-19
No status updates after a few hours (Chamberlain ML1000ev)


michaeljoos72:
Hi all

Recently I have connected a ratgdo board with ESPHome-FW to my Chamberlain ML1000ev Garagedoor opener (“yellow learn button” Security + 2.0). The board I have purchased directly from ratgdo (V2.5) and is powered over USB. WIFI is 3m away from board (-40 dBm).

Connection/Wiring is OK because in general everything works (open/close commands, light, obstruction)

My problem: After a few hours the 5 minutes status reporting stops. This screenshot was taken at 16:30:

![5 Minutes Reporting](https://github.com/ratgdo/esphome-ratgdo/assets/20517474/3d916d05-659e-4728-98c2-a42a2923c9c1)

No more updates and opening the garage door with a remote control also reported no change in status.

"Query status" helped and the status was updated immediately:

![Query Status](https://github.com/ratgdo/esphome-ratgdo/assets/20517474/282e0c35-3415-4b04-9ef0-9c3e6a97c552)

Do you know what could be the root cause for this? Does the Garagedoor opener somehow have a ndeep sleep? And "query status" does wake-up?

Thanks & BR
Michael

tmjo:
Hi,

I think I see the same behavior as you report here. My setup is a Liftmaster LM100EVS with an original Ratgdo v2.5i flashed with latest esphome-version. Everything working fine (except close-command due to not having obstruction sensors ref [here](https://github.com/ratgdo/esphome-ratgdo/issues/146#issuecomment-1886904069), but toggle works perfectly).

I get correct status for a while, and then it stops updating. Seems like toggle-command from webUI or HA still works, and then status update is back again. A query status command also get things back on track.

Any ideas why this might happen?

damaar666:
> Hi,
> 
> I think I see the same behavior as you report here. My setup is a Liftmaster LM100EVS with an original Ratgdo v2.5i flashed with latest esphome-version. Everything working fine (except close-command due to not having obstruction sensors ref [here](https://github.com/ratgdo/esphome-ratgdo/issues/146#issuecomment-1886904069), but toggle works perfectly).
> 
> I get correct status for a while, and then it stops updating. Seems like toggle-command from webUI or HA still works, and then status update is back again. A query status command also get things back on track.
> 
> Any ideas why this might happen?

This is similar to my experience,  with the obstruction sensors.

tmjo:
[...]

PR #607, opened 2026-04-15
Implement a watchdog to query for status

PaulWieland:
If no status updates are received from sec + 2 opener after a timeout, query for status. Ref issue #179

### DIFF
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -140,7 +140,7 @@ namespace secplus2 {
             // Rollover-safe: unsigned subtraction wraps correctly across millis() overflow.
             const uint32_t now = App.get_loop_component_start_time();
             if (static_cast<uint32_t>(now - this->last_status_ms_) > STATUS_WATCHDOG_TIMEOUT) {
-                ESP_LOGW(TAG, "No status received in 11 minutes, querying status");
+                ESP_LOGW(TAG, "No status received in 6 minutes, querying status");
                 this->query_status();
                 this->last_status_ms_ = now;
             }
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -45,6 +45,11 @@ namespace secplus2 {
         this->traits_.set_features(Traits::all());
         this->last_status_ms_ = App.get_loop_component_start_time();
         this->start_status_watchdog();
+
+        this->ratgdo_->set_interval(INTERVAL_PING, PING_INTERVAL, [this]() {
+            ESP_LOGD(TAG, "Sending PING command");
+            this->send_command(Command(CommandType::PING));
+        });
     }
 
     void Secplus2::loop()
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -315,6 +315,17 @@ namespace secplus2 {
                 this->rx_last_read_ = millis();
                 this->rx_packet_[this->rx_byte_count_] = ser_byte;
                 this->rx_byte_count_++;
+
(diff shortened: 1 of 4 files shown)


PaulWieland:
Waiting for user testing and logging. The timeout may need to be adjusted to 5 min, X seconds.

components/ratgdo/secplus2.cpp
```diff
@@ -131,6 +132,15 @@ namespace secplus2 {
         this->sync_helper(millis(), 500, 0);
     }
 
+    void Secplus2::reset_status_watchdog()
```
bdraco: I'd go with set_interval and member var with rollover safe math. `set_interval` is MUCH cheaper to re-arm as `set_timeout` has to create a new one every time it fires.

bdraco (approved):

components/ratgdo/secplus2.cpp
PaulWieland: I wonder if we even need a rollover. Maybe just sending a request every so often is enough to keep the data stream alive.

bdraco: I'd keep the check and rollover guard to minimize collisions

PaulWieland: I was thinking more about making it proactive instead of reactive. If we wait for status updates to stop, then there is a period of time where the door wouldn't report anything (apparently even if its commanded to open or close). By periodically (maybe every 1 hour) pinging the opener with a command, it might keep the data stream alive.

bdraco (approved):

PaulWieland (commented):
Will ask for a long term test and log capture in #179 
If my theory is correct, sending a ping at some interval will prevent the reactive watchdog from ever firing.

PaulWieland:
the pings seemed to be ignored. The packets are however being mangled in some cases a new packet is received while the buffer is partially full from another packet. Why the first packet isn't completely received could possibly be from a buffer overflow, but the firmware also lost the second packet because it began storing it in the middle of the aborted packet's buffer. The latest change attempts to clear the buffer when the packet header is read so at least the second packet isn't lost.

components/ratgdo/secplus2.cpp
```diff
@@ -299,6 +315,17 @@ namespace secplus2 {
...
+
+                this->rx_msg_start_ = ((this->rx_msg_start_ << 8) | ser_byte) & 0xffffff;
+                if (this->rx_msg_start_ == 0x550100) {
+                    ESP_LOGD(TAG, "Sync sequence found mid-packet. Discarding %d bytes and restarting.", this->rx_byte_count_ - 3);
+                    this->rx_packet_[0] = 0x55;
+                    this->rx_packet_[1] = 0x01;
+                    this->rx_packet_[2] = 0x00;
+                    this->rx_byte_count_ = 3;
+                    continue;
+                }
```
bdraco, line 327: Is there any legitimate packet that can have 0x55, 0x01, 0x00 inside of it ?

bdraco: yes it can, but the trade off might be worth it as the probability is very low

bdraco: another approach would be to let decode_wireline fail with -1, and than rewind the buffer if we have  0x55, 0x01, 0x00 mid sequence.

PaulWieland: Results from the testing show an 80% reduction in errored packets and thus watchdog interventions. I'm also not sure that a valid sec + 2.0 packet can coincidentally contain the header sequence. If it could, what good would a header sequence be? Anyway even if it did, I think the chances are ridiculously low. I'm good with it the way it is.

bdraco: It possible, but the odds are roughly ~1-in-a-million packets. So it would take 10 years to hit. Agree its not worth handling as its just a short stall to recover anyways

Issue #633, opened 2026-05-28
Door opener no longer processing commands from Ratgdo32 or v2.5i


Daz42:
This isn't really an issue for Paul to look into since I found the issue - more to give anyone who might be having the same issue a heads up.

A week ago my garage door opener was not responding via my v2.5i. I could see in the real time logging commands being sent and packets being exchanged. I assumed my ratgdo v2.5i was at fault so I ordered 2 new ratgdo32 units from Paul. When they arrived, I installed one and it behaved in exactly the same way. The dry contact wall button worked and the remotes all worked.

So I was initially at a loss. Luckily I had a spare garage door opener and both the old v2.5i and new ratgdo32 units work.

The most likely and weird thing is that there must be something wrong with my original garage door opener. I switched to the dry contact firmware and of course that works fine but no status from the door (open/close) without a reed switch and not getting any feedback from the obstruction sensors. I don't know if anyone else has experienced this - or if somehow Chamberlain have pushed a firmware update to my opener via the MyQ module (seems a bit cynical but you never know with that company).

For now I am using Claude code to implement a dry contact open with all the security+ features (disabling open/close via security plus) - so I don't have to install the reed switches and still get the open/close/lock/obstruction states. Might be helpful with someone in the same situation. It's so weird that this has stopped working while the wall and remotes are all working fine.

PaulWieland:
Why don't you post the ratgdo log so we can figure out why your door opener isn't responding?
Which control protocol is it? 1.0?

Daz42:
> Why don't you post the ratgdo log so we can figure out why your door opener isn't responding? Which control protocol is it? 1.0?

I can definitely do that. I'll flash the Security+ 2.0 firmware and get those logs and post them here. It's Security+ 2.0 and the door opener is a Merlin Commander Elite.

Daz42:
```
Button | OFF |  
Client ID | 8222009 | 1337 8385849
Closing duration | 19.8 s | 0.0 180.0
Door | OPEN | ↑ ☐ ↓
Dry contact close | OFF |  
Dry contact light | OFF |  
Dry contact open | OFF |  
Firmware Version | 2026.5.0 (config hash 0x57879985, built 2026-05-22 19:43:08 +0000) |  
LED | OFF | Off   On
Learn | OFF | Off   On
Light | OFF | Off   On
Lock remotes | UNLOCKED | 🔐 🔓 ↑
Motion | OFF |  
Motor | OFF |  
[...]

PR #637, opened 2026-06-05
Fix startup sync bug

PaulWieland:
increment_rolling_code_counter was never firing because the DoorState is broadcast by the GDO.

Attempt to fix issue #633

### DIFF
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -109,7 +109,7 @@ namespace secplus2 {
             return;
         }
 
-        if (tries == 2 && !synced) { // made a few attempts and no progress
+        if (tries == 2) { // made a few attempts and no progress
             // increment rolling code counter by some amount in case we crashed without writing to flash the latest value
             this->increment_rolling_code_counter(MAX_CODES_WITHOUT_FLASH_WRITE);
         }
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -109,7 +109,7 @@ namespace secplus2 {
             return;
         }
 
-        if (tries == 2 && *this->ratgdo_->door_state == DoorState::UNKNOWN) { // made a few attempts and no progress (door state is the first sync request)
+        if (tries == 2 && !synced) { // made a few attempts and no progress
             // increment rolling code counter by some amount in case we crashed without writing to flash the latest value
             this->increment_rolling_code_counter(MAX_CODES_WITHOUT_FLASH_WRITE);
         }


bdraco:
@bluetoothbot review

components/ratgdo/secplus2.cpp
```diff
@@ -109,7 +109,7 @@ namespace secplus2 {
             return;
         }
 
-        if (tries == 2 && *this->ratgdo_->door_state == DoorState::UNKNOWN) { // made a few attempts and no progress (door state is the first sync request)
+        if (tries == 2 && !synced) { // made a few attempts and no progress
```
bdraco: We will always be synced here because there is a guard on line 108

bluetoothbot: Confirmed. Guard returns when synced. Past line 110, never synced. `!synced` always true. Condition reduces to `tries == 2`. Write that directly. Also broadens trigger beyond door state to any pending query.

bdraco: need to check if door state can flip to DoorState::UNKNOWN when we have had a working reply.  If so maybe we need a DoorState::NONE for when we have had no sync

PaulWieland: I don't think we can use DoorState for this as its set as soon as the GDO decides to broadcast it.

bdraco: So the guard is probably wrong. Probably should be if we ever got back a response to an explicit message that required it to be in sync

PaulWieland: good catch

bdraco: maybe safest guard is leave the existing condition as-is and put an || tries == 3 so we have another fallback

bluetoothbot: Sound fallback. Line 108 guard already forces unsynced here. So `|| tries == 3` fires only when still unsynced. Keeps door-state path. Adds deterministic late bump. This diff instead drops door-state and fires at `tries == 2` unconditionally. Recommend your `|| tries == 3` form. Update stale comment too.

PaulWieland: Hmmm. I was thinking that the existing condition basically _never_ works, but maybe that assumption is incorrect.

PaulWieland: Right, that was my original attempt but I missed the return on 108. Basically if we get back GET OPENINGS or any response from GET STATUS it sync'd.

bdraco: I guess the question is if we can ever get UNKNOWN set. need to check

bdraco: Yeah it can so the unknown check is the wrong signal regardless and the drop is correct

bdraco (approved):

bdraco:
Not sure it will fix https://github.com/ratgdo/esphome-ratgdo/issues/633 though but its still a good cleanup

components/ratgdo/secplus2.cpp
PaulWieland: UNKNOWN is the initial value at boot, and its also set if the GDO ever sends back a value that wasn't mapped.

PaulWieland:
> Not sure it will fix https://github.com/ratgdo/esphome-ratgdo/issues/633 though but its still a good cleanup

Checking the logs in #633, its starting up and getting the door status almost immediately. Because the rolling code is out of sync, the door opener will not respond to commands. (incidentally, it seems like GET_STATUS works even with an old rolling code). This keeps happening because they don't have anything connected to the API, so it reboots every 15 minutes.

PR #644, opened 2026-06-24
Ratgdo encoder sec+

PaulWieland:
Adds encoder support for sec + 1&2. 

For these protocols, the encoder can detect if the door is decoupled from the door opener and manually opened. The encoder should be installed on the overhead door spring pulley so that its reading is coupled with the door itself and not the opener.

The implementation intends to keep the door status from the opener as priority. The status is overidden by the encoder if it detects movement without any status update from the door opener itself. A manually operated sensor is included to indicate when the encoder overrides the status. 

The reverse direction setting is not needed since we can align the encoder direction from the door opener status directly.


PaulWieland:
@bluetoothbot review

bluetoothbot (changes requested):
Blocking issues found.

- YAML refs repointed to feature branch — must revert to `main` before merge

bluetoothbot (changes requested):
Blocking issues found.

- YAML refs pinned to feature branch instead of main

bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
Blocking issues found.

- Shipped configs pinned to feature branch instead of main

bluetoothbot (changes requested):
Blocking issues found.

- Shipped configs pinned to feature branch instead of main

PaulWieland:
@bluetoothbot  review

bluetoothbot (changes requested):
Blocking issues found.

- Grace window is measured from last proto change, not motion onset — normal opener operation can transiently trip manually_operated

PaulWieland:
@bluetoothbot  review

bluetoothbot (changes requested):
Blocking issues found.

PaulWieland:
@bluetoothbot  review

bluetoothbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Manual-operation latch is unreachable — `encoder_received()` is only called once per direction change
> - Regression: drycontact encoder builds lose door state entirely (`protocol_door_state_` is always UNKNOWN)

PaulWieland:
@bluetoothbot  review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

PR #645, opened 2026-06-24
[RFC] Configure inverted obstruction pin logic via platform pin settings

kaechele:
Looking for some comments here on this alternative suggestion to dealing with inverted logic on the obstruction sensor signal.

This essentially reverts #511 and leaves configuring expected logic states to the platform pin configuration and initialization.

Currently the component code does not allow configuring the obstruction input pin pullup, as it is hardcoded to be enabled on ESP32: https://github.com/ratgdo/esphome-ratgdo/blob/77db677ec0572fcb96fa6b423279f59d6ca8ff41/components/ratgdo/ratgdo.cpp#L87

#511 introduced another logic path and configuration option to work around a difference in board design between revisions. To me it makes more sense to handle these difference declaratively (i.e. in the ESPHome configuration) rather than bending the logic in the code around how board revisions are constructed.
This approach simplifies reasoning over the logic inside the component code and ensures that things such as interrupts and pull-ups/pull-downs behave as intended, as they were not part of #511.

On boards that use the old circuitry (voltage divider instead of MOSFET) on the obstruction sensor input keeping the pullup enabled (even though it is not needed) raises the floating voltage of the pin when the sensor signal is low which can cause misreads.

I tested this code on a board that uses the old voltage divider circuit and an ESP32. I have not tested this with a new v32 board (as I don't have such board), so I would be interested if someone could test this and report back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,3 +48,23 @@ Most ESP32 boards use the **ESP-IDF** framework. The project originally depended
 The **v3.2 Disco** board uses the Arduino framework because its VL53L4CX distance sensor library and Wire I2C library require it.
 
 ESP8266 boards continue to use the Arduino framework as ESPHome requires it on that platform.
+
+## Troubleshooting
+
+### False obstruction events
+
+Starting with the v32 board series the obstruction sensor input uses circuitry that causes logic levels to be inverted. If the input pins for the obstruction sensor are not configured correctly this can cause misreads on the obstruction sensor in some instances.
+
+This is accounted for in the configurations provided in this repository. However, if you are creating a custom ESPHome configuration you need to ensure that the pin connected to the obstruction sensor is configured with the `mode` set to `INPUT_PULLUP` and the `inverted` flag set to `true` for these newer boards:
+
+```yaml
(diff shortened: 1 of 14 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
Blocking issues found.

- v32 party-mode example missed the inverted-pin update
- Internal pullup silently dropped for all non-v32 ESP32 boards — only one variant tested
- Removing obst_sleep_low is a breaking config change with no migration note

kaechele:
I added the changes as suggested by the bot in a `fixup!` commit so you can review them individually. That commit can be autosquashed before potentially merging.

bluetoothbot (changes requested):
Blocking issues found.

- Internal pullup dropped for v2 d1_mini / s2_mini ESP32 — untested variants

kaechele:
>     * Internal pullup dropped for v2 d1_mini / s2_mini ESP32 — untested variants

If someone could test with other hardware variants I don't own that'd be fabulous.

Issue #311, opened 2024-07-22
Sec 1.0 discarded packets even on latest release, device still entering emulation mode, still causing door to open and close


jgluck:
Sec 1.0 discarded packets even on latest release, device still entering emulation mode, still causing door to open and close. 

For some reason, occasionally my 889LM takes about 30 seconds to start up. Is there any way to extend the wait before entering emulation mode to 1 minute? Or make it configurable?

<img width="369" alt="image" src="https://github.com/user-attachments/assets/20aa02c6-e0b1-4e54-a305-b7ccc96c7131">

Even now, our door is opening and closing on its own.

jgluck:
Short term I have flashed andy-cooper’s branch from #272 , and we’ll see if it sticks for now. I’m happy to pull any logs necessary. I’m entirely willing to believe that I have a faulty 889LM; however, even so I’d prefer the negative outcome to be the door opener doesn’t respond, instead of randomly opening and closing.

PaulWieland:
I would be interested in seeing the serial log if you can share it. From as close to startup as you can get it.

jgluck:
I’ll try to grab one. @PaulWieland are those just the streaming logs available via the ESPHome plugin (or on the ESPHome landing page?)

PaulWieland:
From the esphome webui on the device itself.

gullygossner:
are there any updates on this issue?

PaulWieland:
Still waiting on a serial log showing it happening.

Issue #226, opened 2024-02-14
Garage door sometimes open when a light command is sent (Security+ 1.0 / 889LM)


cjhesse:
(Moving this from a comment in another issue because the cause was unrelated.)

On three separate occasions, a command to the opener's light resulted in the door opening. Unfortunately, I haven't been able to reproduce the issue while viewing the logs.

I discovered the issue when using an automation to turn the garage door light on when the door to the garage opened. When the automation was triggered:

1. Command the GDO light on.
2. Wait 5 minutes.
3. Command the GDO light off.

The automation runs in "restart" mode, so it will stop the current instance and start over if it's triggered while still running.

This automation was in use with the ratgdo MQTT firmware for about a week with no issues prior to changing to the ESPHome firmware.

In lieu of ratgdo logs, here are the sequence of events in Home Assistant and Frigate (NVR):
```
05:45:34: The door to the garage opens and the automation is triggered.
05:45:34: The video from Frigate shows the GDO light turn on and the garage door opening at effectively the same time.
05:45:34: Home Assistant logs show the GDO light was turned on by the automation.
05:45:37: Home Assistant logs show that the garage door is opening.
05:45:38: Home Assistant logs show that the GDO light was turned off by the automation. Video from Frigate shows that the light **did not** turn off.
05:45:38: Home Assistant logs show that the GDO light was turned on by the automation.
05:45:48: Home Assistant logs show that the garage door was fully opened.
05:45:53: The door is opened again and the automation restarts.
05:46:38: The door is opened again and the automation restarts.
05:51:39: Home Assistant logs show that the GDO light was turned off by the automation. This is correct.
```

Something similar has happened on two other occasions. Once, when the light was being commanded off (at the end of the automation) and once when the light was being commanded on. The sequences for the other two occasions were similar to what's described above, including some incorrect Home Assistant entries about the light turning on or off when it wasn't. For the occasion where the garage door was opened when the light was commanded off, for no amount of time did the light actually turn off.

I've since attempted (unsuccessfully) to reproduce the issue while logging. While I didn't get the door to open, I did capture a few instances of nothing occurring when a command was sent. I'm not sure if these are valuable, but here are a few samples:

```
[20:20:28][D][ratgdo:204]: Light state=OFF
[20:20:28][D][ratgdo:210]: Lock state=UNLOCKED
[20:20:28][D][ratgdo:095]: Door state=CLOSED
[20:20:28][D][light:036]: 'Light' Setting:
[...]

Issue #120, opened 2023-11-21
Light turns on by itself a random times? [2.5i board, ESPHome]


shanelord01:
Module: 2.5i
Firmware: 2023.11.2 (Nov 19 2023, 21:43:41)

I have twi 2.5i boards [correctly wired](https://github.com/PaulWieland/ratgdo/issues/132) to my Merlin 2.0+ door openers.

![IMG_1301](https://github.com/ratgdo/esphome-ratgdo/assets/49434655/d8db032d-69b2-4dd7-9a15-39e9ec25ed51)
![IMG_1302](https://github.com/ratgdo/esphome-ratgdo/assets/49434655/7212a3cd-a273-45f3-9983-0a10c96c1000)

They are then integrated into Home Assistant. At random times the lights turn on bny themselves and this is what the log shows:

```
10:16:25 | [D] | [ratgdo:210] | Status: door=CLOSED light=OFF lock=UNLOCKED
-- | -- | -- | --
10:16:43 | [D] | [light:036] | 'Light' Setting:
10:16:44 | [D] | [number:012] | 'Rolling code counter': Sending state 131.000000
10:16:44 | [D] | [esp8266.preferences:238] | Saving preferences to flash...
10:17:01 | [D] | [ratgdo:210] | Status: door=CLOSED light=ON lock=UNLOCKED
```

Home Assistant looks like this:
![ha-ratgdo](https://github.com/ratgdo/esphome-ratgdo/assets/49434655/28008b08-bb54-4a87-b18a-856f8eb7d3a4)

If it wasn't for my auto-off automation, these lights stay on constantly.

Any ideas?

Thanks,
Shane.

judeibe:
I  having the same symptoms on a Craftsman security 1.0.

finder39:
This is happening for me too on a Chamberlain Liftmaster security 1.0

judeibe:
> This is happening for me too on a Chamberlain Liftmaster security 1.0

I solved this case. Power cycle the unit by unplugging. It will be fine the next restart. Do not turn on the light by pressing the button. If you have a motion sensor on the switch turn it off as well.

finder39:
> > This is happening for me too on a Chamberlain Liftmaster security 1.0
> 
> I solved this case. Power cycle the unit by unplugging. It will be fine the next restart. Do not turn on the light by pressing the button. If you have a motion sensor on the switch turn it off as well.

This seems to work! Thanks!

[...]

PR #647, opened 2026-06-29
Security+1 reliability: reliable light toggle, steady readback, and post-reboot bus sync

fizzsnob:
Three small, related Security+ 1.0 reliability fixes, all in `secplus1`, found and hardware-tested together on a Chamberlain Security+ 1.0 opener with an 889LM-style wall panel (ratgdo32, `v32board_secplusv1`). The commits are independent — take any subset.

## 1. Light toggle nets no change (send a second release)

On some Security+ 1.0 openers, toggling the opener **light** makes it reach the requested state and then **revert ~1–2 s later**, ending unchanged. Reported by several users (e.g. #226, #120, #311).

A device log on an affected opener (one command to turn the light **on** while it was **off**) shows a single press + single release, and the polled `Light state` never changing:

```
Sent byte: [32]      (light press)
Sent byte: [33]      (light release, ~500 ms later)
Light state=OFF      (before AND after — net no change)
```

The current code sends one press + one release ~500 ms apart. The HomeKit ratgdo firmware sends the press followed by **at least two** releases (`set_light` → `sec1_light_release`, `std::max(2, n)`), and a real wall panel transmits the release state continuously. Empirically, a single delayed release does not produce a stable toggle on these openers, while a second release does. The exact opener-side cause isn't established from the logs — this mirrors the proven HomeKit / wall-panel behavior.

## 2. Reported light state bounces on/off/on (suppress stale readback)

Security+ 1.0 has no command ack; light state is learned only by polling `QUERY_OTHER_STATUS`, and there's a ~1 s settle window after a command during which a poll carrying the *old* state bounces the reported state. Record the command time and ignore polled light status for 2 s afterward — record-and-ignore only, it never asserts an unconfirmed state. Only visible once the light actually toggles (fix 1). Gated so it can't suppress readback during the first seconds after boot, and (per review) the window is stamped when the light bytes are **actually transmitted**, not just requested, so a congestion-delayed send can't let it expire early.

## 3. First command after a reboot can collide (wait for wall-panel traffic)

After a reboot the ratgdo joins an already-running bus. Its collision-avoidance keys off `last_rx_`, but that's `0` at boot, so it reads the bus as idle and can inject the first command into the middle of the wall panel's in-progress message, garbling it. Hold transmissions until we've actually heard the panel (`last_rx_ != 0`) — **or** until the bus has stayed silent for 2 s (`BUS_QUIET_GRACE_MS`), at which point no panel is actively transmitting and there is nothing to collide with. A present, running panel is heard within milliseconds, so panel setups key off real traffic; a no-panel setup waits at most ~2 s. *(The first version of this gate held no-panel setups until emulation engaged, ~35 s — caught in review and fixed.)*

## Testing

Hardware-confirmed on a Chamberlain Security+ 1.0 with an 889LM-style wall panel (ratgdo32, `v32board_secplusv1`):

- Fixes 1 + 2: the light now toggles once and stays, and the web UI no longer flickers. Confirmed over repeated toggles.
- Fix 3: addresses a **one-time** light revert seen only on the first command after a reboot (a bus collision before the ratgdo had synced). It's intermittent by nature, so it is reasoned from the boot / `last_rx_` behavior rather than reproduced on demand.
- Review follow-ups (quiet-bus grace, send-time window stamping): compile-verified. The no-panel grace path cannot be exercised on this hardware (a wall panel is present), so it is verified by inspection only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/ratgdo/secplus1.cpp
+++ b/components/ratgdo/secplus1.cpp
@@ -16,6 +16,10 @@ namespace secplus1 {
     using namespace scheduler_ids;
     static const char* const TAG = "ratgdo_secplus1";
     static constexpr uint32_t DOOR_STATE_CALLBACK_TIMEOUT = 2000;
+    // After a light command, ignore polled light status this long so the opener's
+    // reporting latency cannot bounce the reported light state on/off/on while it
+    // settles.
+    static constexpr uint32_t LIGHT_CMD_SUPPRESS_MS = 2000;
 
     void Secplus1::setup(RATGDOComponent* ratgdo, Scheduler* scheduler, InternalGPIOPin* rx_pin, InternalGPIOPin* tx_pin)
     {
@@ -138,6 +142,9 @@ namespace secplus1 {
         }
         if (
             action == LightAction::TOGGLE || (action == LightAction::ON && this->light_state == LightState::OFF) || (action == LightAction::OFF && this->light_state == LightState::ON)) {
+            // Record when the command was issued so QUERY_OTHER_STATUS can
+            // briefly ignore stale readback and not bounce the reported state.
+            this->light_command_at_ = millis();
             this->toggle_light();
         }
     }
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
Blocking issues found.

- Boot-sync gate delays first command by up to ~35s in no-panel setups

fizzsnob:
Thanks @bdraco for triggering the review — both findings were correct. Addressed in fe6a360:

**🟡 Boot-sync gate delayed the first command ~35 s on no-panel setups** — confirmed: nothing sets `last_rx_` without a panel (the GDO only answers polls), and emulation was the only other transmit path. The gate now falls back to transmitting once the bus has stayed **silent for 2 s** (`BUS_QUIET_GRACE_MS`, anchored at listen start in `setup()` and re-anchored by `sync()`): a present, running panel is heard within milliseconds, so panel setups still key off real traffic, while a silent bus waits at most ~2 s — and has nothing on it to collide with. The PR description's incorrect "as before" claim is fixed as well.

**🟢 Suppression window keyed to request time** — the window is now re-stamped in `do_transmit_if_pending()` when a light press **or release** byte is actually popped for transmit (both the loop and wall-panel-emulation paths go through it), keying it to the last light byte on the bus; the request-time stamp is kept so the request→send gap stays covered. On `light_command_at_` never being reset: intentional — the `!= 0` boot guard plus unsigned `millis()` deltas mean the window ages out naturally, and wraparound yields a large delta (window inactive) rather than a stuck window.

Testing note: the original three fixes remain hardware-verified on a Chamberlain Security+ 1.0 with an 889LM-style wall panel. The follow-up commit is compile-verified; the no-panel grace path can't be exercised on this hardware (a panel is present), so that path is verified by inspection only.

bluetoothbot (approved):
No blocking issues found.

PR #649, opened 2026-07-04
Fix format specifiers in secplus2 logging

bdraco:
The ESP-IDF build warns about %d used with uint32_t and uint64_t arguments in secplus2.cpp; the Client ID line printed a truncated value on 64 bit arguments.

Use PRIu32 and PRIu64, matching the PRIx32 and PRIx64 usage already in the file; combine the dump_config lines into a single ESP_LOGCONFIG call per https://developers.esphome.io/architecture/logging/#the-solution-combine-related-log-messages

The remaining Wvolatile warnings from ratgdo.h are a build system issue, handled upstream in esphome.

### DIFF
--- a/components/ratgdo/secplus2.cpp
+++ b/components/ratgdo/secplus2.cpp
@@ -63,9 +63,11 @@ namespace secplus2 {
 
     void Secplus2::dump_config()
     {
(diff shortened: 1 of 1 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #672, opened 2026-08-17
Fix -Wformat mismatches for millis() in log format strings

martinemde:
Hi, thanks for making this. I just ordered a second non-disco for my side garage door since it works so much better than the old way. Here's a small fix for warnings I was seeing during compile. Flashed to my disco, it works as expected.

## What

Log statements that print `millis()` used `%d` and `%ld` conversions that do not match the argument type. This switches them to `PRIu32` and adds the `<cinttypes>` include where it was missing.

## Why

`millis()` returns `uint32_t`, which both the ESP32 and ESP8266 toolchains define as `long unsigned int`. Two of these calls, in `secplus1.cpp` at the `Discard incomplete packet` warning and the `Sent byte` debug line, sit on unconditional paths and emit `-Wformat` warnings on every board build. The rest are behind `ESP_LOG1` and `ESP_LOG2`, so they stay quiet until the logger level is raised to `very_verbose`, but they are wrong in the same way.

## How

Replace the mismatched conversions with `PRIu32` in `ratgdo.cpp`, `secplus1.cpp`, and `secplus2.cpp`. `secplus2.cpp` already used `PRIx32` and `PRIx64` through a transitive include; the explicit `<cinttypes>` there makes that dependency real rather than incidental. Format strings only, no behaviour change.

## Testing

`pytest tests/ -v` passes (15 tests), and `pre-commit run --files` passes on all three changed files, including `clang-format -style=Webkit`.

Flashed to a real ratgdo32 Disco driving a Security+ 1.0 opener with a digital wall panel. The `v32disco_secplusv1` build now compiles with zero warnings, where the two unconditional ones appeared before. The board ran clean after flashing with healthy Sec+1.0 packet decode, and the garage light was toggled from Home Assistant to exercise the transmit path that the `Sent byte` line logs from.

<details>
<summary>Agent run details (optional, for reviewers)</summary>

Pytest: `python -m pytest tests/ -q`, 15 passed.

Pre-commit: `pre-commit run --files components/ratgdo/ratgdo.cpp components/ratgdo/secplus1.cpp components/ratgdo/secplus2.cpp`. trailing-whitespace, end-of-file-fixer, check-added-large-files, and clang-format all passed with no rewrites.

Hardware: 169 seconds of continuous log capture from the flashed board showed one boot banner and no repeats, ruling out a reboot loop. `Light state=OFF` and `Lock state=UNLOCKED` were each logged 337 times over that window from `RATGDOComponent::received(...)`, roughly two per second, confirming RX and packet decode are healthy on the patched firmware. WiFi held at -34 dBm with no errors.

Type note: `RATGDOComponent::door_start_moving` is declared `unsigned long` rather than `uint32_t`. On both target toolchains those are the same type, which is why the old `%d` warned there, so `PRIu32` matches. Retyping the member was left out to keep this to format strings.

</details>

Drafted with Claude Code (Opus 5); reviewed and hardware-tested by @martinemde (yes actually).

### DIFF
--- a/components/ratgdo/ratgdo.cpp
+++ b/components/ratgdo/ratgdo.cpp
@@ -15,6 +15,8 @@
 #include "common.h"
 #include "ratgdo_state.h"
 
+#include <cinttypes>
+
 #ifdef PROTOCOL_DRYCONTACT
 #include "dry_contact.h"
 #endif
@@ -526,7 +528,7 @@ void RATGDOComponent::schedule_door_position_sync(float update_period)
 {
     ESP_LOG1(
         TAG,
-        "Schedule position sync: delta %f, start position: %f, start moving: %d",
+        "Schedule position sync: delta %f, start position: %f, start moving: %" PRIu32,
         this->door_move_delta, this->door_start_position,
         this->door_start_moving);
     auto duration = this->door_move_delta > 0 ? *this->opening_duration
@@ -557,7 +559,7 @@ void RATGDOComponent::door_position_update()
         return;
     }
     auto position = this->door_start_position + (now - this->door_start_moving) / (1000 * duration);
-    ESP_LOG2(TAG, "[%d] Position update: %f", now, position);
+    ESP_LOG2(TAG, "[%" PRIu32 "] Position update: %f", now, position);
     this->door_position = clamp(position, 0.0f, 1.0f);
 }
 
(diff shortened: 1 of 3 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

PR #681, opened 2026-09-01
Add prefer encoder door status switch

hoveeman:
## What

Add a persistent prefer_encoder_status switch and expose reverse_encoder across all encoder-enabled base YAMLs (Security+ 1.0, Security+ 2.0, and Dry Contact).

## Why

In Security+ 1.0 and 2.0 encoder configurations, door status resolution is currently hardcoded to prioritize the wall panel/opener wire protocol over the physical encoder. Users with encoder hardware installed who want door status driven primarily by physical encoder rotation can now toggle this preference directly in the web UI and Home Assistant.

## How

- Added prefer_encoder_status switch type to RATGDOSwitch backed by NVS flash preferences.
- Updated RATGDOComponent state arbitration so that when enabled, encoder motion and limit stops directly update resolved door state, wire protocol status updates do not overwrite it during travel, and initial state on boot is properly established from opener sync.
- Added immediate encoder state recalculation when the switch is toggled ON.
- Exposed reverse_encoder and prefer_encoder_status switches in base_secplusv1_enc.yaml, base_enc.yaml, and base_drycontact_enc.yaml, grouped as Encoder Prefer Door Status and Encoder Reverse.

## Testing

- Ran pytest suite (pytest tests/ -v, 16/16 passed).
- Validated ESPHome configuration on all encoder board targets (v32board_secplusv1_enc.yaml, v32board_enc.yaml, v32board_drycontact_enc.yaml, v25iboard_secplusv1_enc.yaml, v25iboard_enc.yaml, v25iboard_drycontact_enc.yaml, v32disco_secplusv1_enc.yaml, v32disco_enc.yaml, v32disco_drycontact_enc.yaml).
- Hardware testing: Flashed to ratgdo v3.2 board with Security+ 1.0 opener and encoder. Verified door open/close operations, reboot state sync, and switch toggling behavior on live hardware.

### DIFF
--- a/base_drycontact_enc.yaml
+++ b/base_drycontact_enc.yaml
@@ -91,6 +91,11 @@ switch:
     type: reverse_encoder
     name: "Encoder Reverse"
     entity_category: config
+  - platform: ratgdo
+    id: ${id_prefix}_prefer_encoder_status
+    type: prefer_encoder_status
+    name: "Prefer Encoder Door Status"
+    entity_category: config
 
 button:
   - platform: restart
--- a/base_drycontact_enc.yaml
+++ b/base_drycontact_enc.yaml
@@ -94,7 +94,7 @@ switch:
   - platform: ratgdo
     id: ${id_prefix}_prefer_encoder_status
     type: prefer_encoder_status
-    name: "Prefer Encoder Door Status"
+    name: "Encoder Prefer Door Status"
     entity_category: config
 
 button:
--- a/base_enc.yaml
+++ b/base_enc.yaml
@@ -113,6 +113,16 @@ switch:
     name: "Learn"
     icon: mdi:plus-box
     entity_category: config
+  - platform: ratgdo
+    id: ${id_prefix}_reverse_encoder
+    type: reverse_encoder
+    name: "Encoder Reverse"
+    entity_category: config
+  - platform: ratgdo
+    id: ${id_prefix}_prefer_encoder_status
+    type: prefer_encoder_status
+    name: "Prefer Encoder Door Status"
+    entity_category: config
 
 binary_sensor:
   - platform: ratgdo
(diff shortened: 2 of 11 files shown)


PaulWieland:
Why is this necessary? If you use this then you will not have the security notification that the door was manually operated. 

If you want to use the encoder for door status on security +1 because of an incompatible wall control panel or other complication, then use dry contact control.

hoveeman:
> Why is this necessary? If you use this then you will not have the security notification that the door was manually operated.
> 
> If you want to use the encoder for door status on security +1 because of an incompatible wall control panel or other complication, then use dry contact control.

With Dry contact control I lost Motion, obstruction detection, light control from the wall panel that works well in Security 1.0. By adding this I can use security 1.0 features and the encoder to know the door status in a reliable state. To your point about manual door opening, I don't have that feature in dry contact either.  

You're welcome to reject this, but perhaps someone else would find it useful as well.

PaulWieland:
Yeah I'm not seeing a real benefit here. If the security+ protocol is working well with your wall control, there is no reason to sacrifice the encoder's input signal to detect that the door was manually operated.

hoveeman:
> Yeah I'm not seeing a real benefit here. If the security+ protocol is working well with your wall control, there is no reason to sacrifice the encoder's input signal to detect that the door was manually operated.

The wall control reports light, obstruction, and motion well. However it doesn't report door status well. It will report open properly, but will flip every 5 seconds from closed to closing and back to closed.

PaulWieland:
There is some sort of fault with the opener then. Have you tried disconnecting your wall panel and using emulation mode to see if the behavior is the same? Have you tried tweaking the door's physical travel limits?

hoveeman:
I just recalibrated the door and there was no difference. I did do emulation mode with the wall panel disconnected and that would work reliably. Would you suggest just putting a dumb doorbell there replacing the wall panel? what is the purpose of the encoder option for security 1.0 and 2.0 when flashing the firmware. It makes sense for dry contact closure. Would this feature be beneficial for others with 1.0 and 2.0?

